PR #14, opened 2021-03-13
Experimental FpVector rewrite

dalcde:
The main problem this is intended to solve is to let us take slices
without having a mutable reference. This should clean up a lot of
existing code. Ideally, this would also lead to cleaner code that is
just as performant. A likely consequence is that we would no longer have
to import the trait FpVectorT to use FpVector, but this is only because
enum_dispatch will not support the new format and we would have to do it
by hand.

The plan is to use const generics to define a FpVectorP type, which is
na FpVector for a fixed P. This allows for monomorphization for any
fixed P without code duplication and type meddling. We then use an enum
to manually wrap the type for various possible values of P and do
enum_dispatch like wrappers. This would also allow us to not hide
internal API's since we are not using a trait.

An FpVectorP is always unsliced. We then introduce a new FpVectorSliceP
type, which represents a slice of FpVectorP, and do the same
dispatching. Most arithmetic functions will be on FpVectorSliceP.


hoodmane:
I think we agreed before that this was a better design that would take more effort to implement. I chose the current way of doing it despite it's inferiority because I was already spending enough time wrestling with my newbie understanding of Rust. Anyways thanks for working on this.

On a related note, I was looking at the disassembly for FpVector addition recently, and it looked to me like the `limb_mask` call in the main loop of the addition was not being inlined for some reason despite the fact that I believe `limb_mask` is marked as `[inline(always)]`. In fact, in the disassembly out of a total of 3990 instructions, there are 139 function calls:
* 98 calls to `panic_bounds_check`
* 6 calls to `panic`
* 31 calls to `limb_mask`
* AddShiftNoneData::new, rust_dealloc, drop_in_place, _Unwind_Resume

I tried compiling without the `odd-primes` feature but the resulting disassembly was identical. Maybe I'm doing something wrong...
Anyways, I'm not sure how many of those calls are actually on execution paths that matter. Anyways if there is a call to `limb_mask` in the main loop of the `add` function, `limb_mask` would probably show up in profiles, so we should double check that. I assume none of the other function calls are inside of a loop and so they don't matter.

Also, no sign of any auto vectorization or simd instructions, here or in the `limb_mask` call. So I think there is definitely a lot of room for improvement.

hoodmane:
I think the addition would be faster if done in two loops: One loop to do the arithmetic, a second loop to reduce. The reducing loop doesn't need to care whether we shift_left, shift_right, or shift_none and doesn't have to worry about partial limbs so it can be shared everywhere.

dalcde:
On Sat, Mar 13, 2021 at 09:34:41AM -0800, Hood Chatham wrote:
>I think we agreed before that this was a better design that would take more effort to implement. I chose the current way of doing it despite it's inferiority because I was already spending enough time wrestling with my newbie understanding of Rust. Anyways thanks for working on this.

I've had numerous attempts at this before and this is the first time the
things managed to come together. I hope this pans out. (the matrix part
of the story may cause issues)

>I tried compiling without the `odd-primes` feature but the resulting disassembly was identical. Maybe I'm doing something wrong...
When did you do this? Until recently disabling odd-primes never worked,
and in any case I've had difficulty disabling it with cargo-asm (I
*know* it was disabled because I get compiler warnings and the
corresponding speed improvement)

>Also, no sign of any auto vectorization or simd instructions, here or in the `limb_mask` call. So I think there is definitely a lot of room for improvement.

Yep.

hoodmane:
> I *know* it was disabled because I get compiler warnings and the corresponding speed improvement

That's right, I see the same effects. What's weird is that to try to get it to work with `cargo asm` I changed the default features in `Cargo.toml` to none, cleaned all the `target` build artifacts, and tried again but I *still* saw exactly the same asm...

dalcde:
I should add that it is easy to see that it's not "the compiler forgot to optimize. When odd-primes is disabled a ValidPrime struct is empty while with odd-primes it is a transparent u32

dalcde:
Another potential optimization is to do the enum_dispatch magic on
matrices as well. Currently when we do matrix operations we have to
check the prime of each row separately, when we know a priori that they
are all the same.

dalcde:
One potential issue with this approach is that this will lead to code
size bloat, since we have one version of the code for every single
prime. The way we do enums means that we need to load all the code even
for primes we don't use, and we could run into issues with L1
instruction cache.

That said, I mostly run my code with odd-primes disabled, so it wouldn't
be a huge issue.

hoodmane:
Could we have a top level function that does dispatch into a distinct function for each prime and make sure that the per-prime functions aren't inlined?

dalcde:
We could make sure that the per-prime functions are't inlined; that may
or may not be helpful depending on how the code is arranged. It would
also make it slower if we don't enable odd-primes.

One possibility is to have a separate feature called "large-primes",
which enables primes > 10.

dalcde:
Progress report: Everything in ext and rust_webserver are working, apart
from the PolynomialAlgebra business. At low primes this is faster and at
large parimes they appear to be comparable. This may be because I
sprinkled in some optimizations as I refactored.

As part of the API change, all the _pure functions were removed; they
are now distinguished by functions on FpVector objects vs functions on
Slices, so the type system forces us to use them correctly.

The odd-primes feature is not yet implemented (as in you get odd primes
regardless). I have thoughts as to how to implement this.

Only primes up to 7 are supported; it is trivially easy to support more
primes though.

@hoodmane I believe everything in python_ext is currently broken. I'm
not very familiar with the code there. How should I proceed?

hoodmane:
I'm okay with you merging it and breaking `python_ext` for now if you want. Hopefully at some point I can go in and fix it.

dalcde:
The performance gains when `odd-primes` is disabled appears significant. One difference in implementation is that instead of making an enum with a single variant, I reexport the equivalent of `FpVector2` as `FpVector`.

PR #20, opened 2021-03-25
Naive SIMD implementation

JoeyBF:
I adapted the code directly from what I wrote a few months ago. For some mysterious reason the AVX2 performance wasn't so bad when I tried it again just now.

Something to keep in mind was that I used a bunch of `loadu` and `storeu` because I found the memory was not aligned in general. We could switch to `load` and `store` if the start of every limb vector was 32B aligned for AVX/2 and 64B for AVX512. The aligned and unaligned instructions should have the same latency and throughput according to the Intel documentation, but that might be something worth investigating.


ext/crates/fp/src/vector_inner.rs
```diff
@@ -9,6 +9,9 @@ use std::cmp::Ordering;
 use std::convert::TryInto;
 use std::sync::Once;
 
+#[cfg(target_feature = "avx")]
```
dalcde, line 12: We would also have to guard this on the target architecture being x86_64.

ext/crates/fp/src/vector_inner.rs
```diff
@@ -245,8 +248,71 @@ impl<const P: u32> FpVectorP<P> {
...
+                    } else if #[cfg(target_feature = "avx")] {
+                        let max_limb = self.limbs.len();
+                        let mut target_limbs_ptr = self.limbs.as_mut_ptr();
+                        let other_limbs_ptr = other.limbs.as_ptr();
+                        let chunks = self.limbs.len() / 4;
+                        for i in 0..chunks {
+                            unsafe {
+                                // SIMD magic
+                                // We need to treat the limbs as floats because packed ints are only supported on AVX2.
+                                // Since we're only XOR'ing and not doing any arithmetic, this is fine.
```
dalcde, line 299: Does this incur any performance penalties? If not we could skip the avx2 branch

ext/crates/fp/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 #![feature(const_panic)]
 #![allow(clippy::many_single_char_names)]
 #![allow(clippy::unreadable_literal)]
+#![feature(stdsimd)]
```
dalcde, line 4: I believe `stdsimd` is already stabilized, so this is not needed.
```suggestion
```

dalcde (commented):
I've messed with trying to allocate more aligned FpVectors, which works, but the higher alignment come with a performance cost on the allocation step. That said, the initial allocation is amortized so this might be a good idea.

dalcde:
I should also add that I can measure a ~10% difference when resolving up to 80, using avx2

ext/crates/fp/src/vector_inner.rs
```diff
@@ -245,8 +248,71 @@ impl<const P: u32> FpVectorP<P> {
...
+                                let mut target_chunk = x86_64::_mm256_loadu_ps(target_limbs_ptr.add(4 * i) as *const float);
+                                let other_chunk = x86_64::_mm256_loadu_ps(other_limbs_ptr.add(4 * i) as *const float);
+                                target_chunk = x86_64::_mm256_xor_ps(target_chunk, other_chunk);
+                                x86_64::_mm256_storeu_ps(target_limbs_ptr.add(4 * i) as *mut x86_64::__m256i, target_chunk);
```
dalcde, line 303: This needs a bit of fixing
```suggestion
                                let mut target_chunk = x86_64::_mm256_loadu_pd(target_limbs_ptr.add(4 * i) as *const f64);
                                let other_chunk = x86_64::_mm256_loadu_pd(other_limbs_ptr.add(4 * i) as *const f64);
                                target_chunk = x86_64::_mm256_xor_ps(target_chunk, other_chunk);
                                x86_64::_mm256_storeu_pd(target_limbs_ptr.add(4 * i) as *mut f64, target_chunk);
```

ext/crates/fp/src/vector_inner.rs
```diff
@@ -245,8 +248,71 @@ impl<const P: u32> FpVectorP<P> {
...
+                            }
+                        }
+                        for i in (4 * chunks)..max_limb {
+                            unsafe { // pointer arithmetic
+                                *target_limbs_ptr.add(i) = *target_limbs_ptr.add(i) ^ *other_limbs_ptr.add(i);
+                            }
+                        }
+                    } else if #[cfg(target_feature = "avx")] {
+                        let max_limb = self.limbs.len();
+                        let mut target_limbs_ptr = self.limbs.as_mut_ptr();
```
dalcde, line 292:
```suggestion
                        let target_limbs_ptr = self.limbs.as_mut_ptr();
```

ext/crates/fp/src/vector_inner.rs
```diff
@@ -245,8 +248,71 @@ impl<const P: u32> FpVectorP<P> {
...
+                            }
+                        }
+                        for i in (8 * chunks)..max_limb {
+                            unsafe { // pointer arithmetic
+                                *target_limbs_ptr.add(i) = *target_limbs_ptr.add(i) ^ *other_limbs_ptr.add(i);
+                            }
+                        }
+                    } else if #[cfg(target_feature = "avx2")] {
+                        let max_limb = self.limbs.len();
+                        let mut target_limbs_ptr = self.limbs.as_mut_ptr();
```
dalcde, line 273:
```suggestion
                        let target_limbs_ptr = self.limbs.as_mut_ptr();
```

ext/crates/fp/src/vector_inner.rs
```diff
@@ -245,8 +248,71 @@ impl<const P: u32> FpVectorP<P> {
         debug_assert_eq!(self.dimension(), other.dimension());
         if P == 2 {
             if c != 0 {
-                for (left, right) in self.limbs.iter_mut().zip(&other.limbs) {
-                    *left = limb::add::<P>(*left, *right, 1);
+                cfg_if::cfg_if! {
+                    if #[cfg(target_feature = "avx512f")] {
+                        let max_limb = self.limbs.len();
+                        let mut target_limbs_ptr = self.limbs.as_mut_ptr();
```
dalcde, line 254:
```suggestion
                        let target_limbs_ptr = self.limbs.as_mut_ptr();
```

dalcde:
This is not going to cut any dependencies so it doesn't really matter, but we don't really need to use `cfg-if`; Raw `if cfg!(...)` works.

The difference is that in the latter, all the branches must be valid code. In our case all the simd primitives are always defined even if they don't make sense on the actual CPU, so the code will compile.

ext/crates/fp/src/lib.rs
JoeyBF: I thought so too but AVX512 is still experimental. I realized that when I tested the code on the Wayne grid.

ext/crates/fp/src/vector_inner.rs
JoeyBF: I thought the AVX instruction sets were only available on x86_64, but you're right that we should make sure. It might change in the future.

ext/crates/fp/src/vector_inner.rs
JoeyBF: I can't tell on my machine right now, the runtimes vary wildly (as in, anything from 20-30 seconds, to 50 or 60, to 2 full minutes). I think it might be because other processes pollute the L1 cache, so if the matrix already barely fits we might end up with more cache misses if the scheduler allows other processes to run on the same core.

ext/crates/fp/src/vector_inner.rs
JoeyBF: Thank you! I should really test all features before committing.

JoeyBF:
> This is not going to cut any dependencies so it doesn't really matter, but we don't really need to use `cfg-if`; Raw `if cfg!(...)` works.
> 
> The difference is that in the latter, all the branches must be valid code. In our case all the simd primitives are always defined even if they don't make sense on the actual CPU, so the code will compile.

I think I was worried that llvm might not optimize it away, so I put the check as a macro before the compilation. Testing shows that it makes no difference, so I removed it.

JoeyBF:
Whoops, of course guarding the import of x86_64 and removing cfg-if are incompatible, even though the avx branches would never actually run.

JoeyBF:
> > This is not going to cut any dependencies so it doesn't really matter, but we don't really need to use `cfg-if`; Raw `if cfg!(...)` works.
> > The difference is that in the latter, all the branches must be valid code. In our case all the simd primitives are always defined even if they don't make sense on the actual CPU, so the code will compile.
> 
> I think I was worried that llvm might not optimize it away, so I put the check as a macro before the compilation. Testing shows that it makes no difference, so I removed it.

If I'm reading the failed checks' logs right, I think core::arch::x86_64 isn't even defined on any other architecture. I think it will be simpler to just stick with cfg-if.

ext/crates/fp/src/lib.rs
dalcde: That makes sense. I believe you can compile to AVX-512 even if your computer doesn't support it.

dalcde:
I'm happy with everything. Good to merge?

JoeyBF:
All good for me!

dalcde:
Thanks!

PR #34, opened 2021-06-03
Save quasi-inverses to disk

### LINKED ISSUE #29: Support saving quasi-inverses on disk
Currently, we store all quasi-inverses in memory, which takes up the bulk of the memory consumption (e.g. up to the 140th stem, we need 25.8 MiB to store the differentials and 40GiB for the quasi-inverses). If we want to resolve to larger degrees, we will run out of memory on any reasonable system.

Since these quasi-inverses are used at very predictable times, we ought to be able store these on the disk and call them on demand. Perhaps a custom allocator could be used for this purpose.

JoeyBF:
Resolves #29

I implemented a new smart pointer that deallocates and reloads wrapped data on demand. I used it to wrap the `preimage` field of `QuasiInverse`, and the rest of the codebase is essentially untouched. I hope the documentation is relatively easy to understand.

I also made some miscellaneous modifications, like fixing typos, adding the option to load from a file in resolve.rs (and changing the tests accordingly) and ignoring some more folders in .gitignore, which is where I run some tests and save my resolutions.

This doesn't pass the check because the `getrandom` package doesn't support wasm. I'm not sure how to fix that. One option would be to forego the `tempfile` package entirely and write our own. It might not be too hard to just rip out the part that we need, since we're already using my personal fork as the source.

Also note that the memory usage is reduced but not as drastically as we hoped. I found that, once the quasi-inverses are cached, the vast majority of the memory is used up by `multiplication_table` in the `AdemAlgebra`. I tried to also wrap the individual `Vec<Vec<FpVector>>`s, and the memory usage does drop to only a few MB, but the computation becomes unreasonably slow (takes several minutes to resolve to t = 60 even with `--resolve` and `--features concurrent` with 6 threads). Of course this is because the vast majority of the time is spent in disk i/o while computing the matrices of `current_chain_map` and `current_differential` in `step_resolution`. I'm not sure how to solve that either.


dalcde:
What does the memory situation look like with the Milnor basis?

JoeyBF:
I just ran two computations with s = 72, t = 100. The default Adem basis used 1135 MB of RAM at its peak, which is right before the end. The Milnor basis used slightly less at 1040 MB. However, both computations started out using much less, around 100 MB (as far as I know), and it grew from there.

I'm not sure what to make of that data.

dalcde:
On Wed, Jun 02, 2021 at 10:44:34PM -0700, Joey Beauvais-Feisthauer wrote:
>I just ran two computations with s = 72, t = 100. The default Adem basis used 1135 MB of RAM at its peak, which is right before the end. The Milnor basis used slightly less at 1040 MB. However, both computations started out using much less, around 100 MB (as far as I know), and it grew from there.
>
>I'm not sure what to make of that data.

One reasonable thing to try is to use a different allocator, e.g. jemalloc.

JoeyBF:
> One reasonable thing to try is to use a different allocator, e.g. jemalloc.

Do you think it's an allocator problem? I get this behavior for both my laptop and the Wayne grid. I just tried switching to jemalloc and the memory usage is roughly similar, around 1.5 GB for Adem and Milnor. Are you saying that you're getting different numbers when you run the same test on your configuration?

The reason I'm thinking the allocator is not the problem is that I tried saving the multiplication tables to disk and they really did take that much space.

dalcde:
One thing I observe (without this PR) is that if we load a saved resolution the memory usage is less than if we compute from scratch, which is perplexing

JoeyBF:
I also get that on my end. Loading from t = 90, Adem takes roughly 1.2 GB and Milnor 1.1 GB.

dalcde:
I think the recent changes to the save format renders this obsolete. Should we
close this?

JoeyBF:
I agree, it would be pointless to merge this since the problem is already solved. Plus the solution was not terribly elegant.

PR #36, opened 2021-09-16
`fp` reorganizing

JoeyBF:
I reorganized some of the code in `fp` in a way that I feel is more coherent. I also added some light documentation, namely for the various limb functions and for constants.

- Most constants (numbers and non-nested arrays) are now defined at build time using a build script. This doesn't change that much but it helped me understand what their value actually is. They are imported into `constants.rs` where `const_for` and other arrays are defined.
- Limb functions are now in their own file instead of `vector_inner.rs` (if I recall correctly).
- Fixed some `u64` hardcoding in the `simd` submodule.
- Introduced criterion and iai for benchmarking `fp` and `ext` itself. Criterion automates the statistical analysis and gives nice graphs, and iai is essentially a wrapper around cachegrind.
- Removed `LIMB_BIT_INDEX_TABLE` for simplicity. My benchmarks tended to indicate that performance was unchanged, but I would be happy to reintroduce it if your benchmarks disagree.


ext/crates/algebra/src/algebra/adem_algebra.rs
```diff
@@ -470,7 +470,7 @@ impl GeneratedAlgebra for AdemAlgebra {
 impl AdemAlgebra {
     pub fn new(p: ValidPrime, generic: bool, unstable: bool, unstable_enabled: bool) -> Self {
         assert!(unstable_enabled || !unstable);
-        fp::vector::initialize_limb_bit_index_table(p);
+        // fp::vector::initialize_limb_bit_index_table(p);
```
hoodmane, line 473: It's better to remove code rather than comment it out, here and in other places.
```suggestion
```

ext/crates/algebra/src/algebra/combinatorics.rs
```diff
@@ -1,10 +1,11 @@
-use fp::prime::*;
+// use fp::prime::*;
```
hoodmane, line 1: Again remove please (etc)
```suggestion

```

ext/crates/fp/benches/multinomial.rs
```diff
@@ -1,29 +1,28 @@
-use bencher::{benchmark_group, Bencher};
+// use criterion::Criterion;
```
hoodmane, line 1: Just delete the whole file?

dalcde:
Most of the changes make sense to me. I think we might as well remove
the initialize_limb_bit_index_table code instead of commenting it out.

I'll do a more detailed review later.

ext/crates/fp/src/limb.rs
```diff
@@ -0,0 +1,221 @@
...
+    pub(crate) limb: usize,
+    pub(crate) bit_index: usize,
+}
+
+/// Return the number of bits an element of $\mathbb{F}_P$ occupies in a limb.
+pub(crate) const fn bit_length<const P: u32>() -> usize {
+    BIT_LENGTHS[PRIME_TO_INDEX_MAP[P as usize]]
+}
+
+/// TODO: Would it be simpler to just compute this at "runtime"? It's going to be inlined anyway.
```
hoodmane, line 22: I agree that this would probably be an improvement.

ext/crates/fp/src/limb.rs
```diff
@@ -0,0 +1,221 @@
...
+            // let prime_idx = PRIME_TO_INDEX_MAP[*p as usize];
+            // debug_assert!(idx < MAX_LEN);
+            // unsafe {
+            //     let table = &LIMB_BIT_INDEX_TABLE[prime_idx];
+            //     debug_assert!(table.is_some());
+            //     *table
+            //         .as_ref()
+            //         .unwrap_or_else(|| std::hint::unreachable_unchecked())
+            //         .get_unchecked(idx)
+            // }
```
hoodmane, line 57: Again, please remove commented code.
```suggestion

```

ext/crates/fp/src/vector_inner.rs
```diff
@@ -690,34 +480,37 @@ impl<'a, const P: u32> SliceP<'a, P> {
...
-        let bit_length = bit_length(self.prime());
-        let entries_per_limb = entries_per_limb(self.prime());
+        let bit_length = limb::bit_length::<P>();
+        let entries_per_limb = limb::entries_per_limb_const::<P>();
         (self.start % entries_per_limb) * bit_length
     }
 
     #[inline]
-    fn limb_range(&self) -> (usize, usize) {
+    fn limb_range(&self) -> Range<usize> {
```
hoodmane, line 489: This change seems to be responsible for a large part of the diff in this file and it seems orthogonal the changes in the rest of the files. Maybe you could do this as a separate PR? (It's certainly a nice improvement.)

hoodmane (commented):
Thanks! These look like nice improvements to me. Please remove all commented code.

These change set is a bit large and it seems to include a lot of orthogonal changes. It would be easier to review if you would make smaller pull requests. For instance, if you split off the `limb_range` return type change, I think that would reduce the diff on this a lot.

JoeyBF:
Ah you're right, I forgot to mention I changed the return type. I'll revert the change and open a new PR.

I commented out the code instead of removing it just in case we wanted to add back `LIMB_BIT_INDEX_TABLE` or the former benches. In an ideal world we could just rely on version control, but I have a bad habit of making commits that are too large. I'll remove the comments in a new commit, so if we want to add them back in I'll just revert that.

JoeyBF:
Also I saw that I added a `_nosimd` version of addition for vectors, for benchmarking purposes. That's also kind of orthogonal, so I could remove it too. I would also remove the commented code in the benchmarks that refers to it.

ext/crates/fp/src/vector_inner.rs
JoeyBF: Thanks for the review! I applied the changes you suggested. Should I "resolve conversation" for your comments above? I'm not sure what the GitHub etiquette is.

dalcde:
I think if you have unambiguously resolved a comment, then you can mark
it as resolved, which is the case here.

dalcde:
It is currently a const function with no arguments. Rust with
monomorphize it for each P. I can't imagine this not being precomputed
and inlined. We should make sure it does that even for debug builds
though, or else the performance impact can be huge.

ext/crates/fp/src/limb.rs
JoeyBF: So I pushed those changes. I waited a while before pushing because I wanted to benchmark thoroughly, but it looks like my setup doesn't want to give me consistent measurements. In any case I don't detect any systematic drop in performance.

One benchmark would seem to indicate that computing the values at runtime actually improves performance at odd primes. If that's really the case, my guess would be that's because LLVM is able to perform a few more optimizations after inlining the functions, because the values are not black boxes anymore.

JoeyBF:
Also by the way, the rust_ext / webserver CI test is a false negative, the server had network problems apparently. The chart build test is the same error as always

PR #40, opened 2021-10-11
Reorganize repository

dalcde:
Feel free to bikeshed on naming and correct inaccurate descriptions.


README.md
```diff
@@ -1,5 +1,36 @@
...
+1. `ext`
+A general library to work with Ext over an Fp algebra. More generally, it
+allows us to work compute in the derived category of said algebra. The primary
+purpose is to compute the classical Adams E2 page by computing Ext over the
+Steenrod algebra.
+
+2. `web_ext`
+Web interfaces to `ext`. There are two subprojects at the moment:
+
+ - `rust_webserver`: A GUI to work with the Adams spectral sequence. Given a
```
mcy, line 15: Do you think naming the webapps after their paths makes sense? E.g. this would be `sseq`? (I just think `rust_webserver` is not particularly suggestive of what the project is, since it's more webapp than webserver.)

web_ext/rust_webserver/Cargo.toml
```diff
@@ -7,15 +7,15 @@ edition = "2018"
...
-algebra = { path = "../ext/crates/algebra", default-features = false }
-bivec = { path = "../ext/crates/bivec", features = ["json"] }
-chart = { path = "../ext/crates/chart" }
-error = { path = "../ext/crates/error" }
-ext = { path = "../ext", default-features = false }
-fp = { path = "../ext/crates/fp", features = ["json"], default-features = false }
-once = { path = "../ext/crates/once" }
-sseq = { path = "../ext/crates/sseq" }
-thread-token = { path = "../ext/crates/thread-token", optional = true }
+algebra = { path = "../../ext/crates/algebra", default-features = false }
```
mcy, line 10: It probably makes sense to reexport these from the `ext` crate, which means we only have one places where we fuss around with dependencies for the helper crates. That will probably be more work and worth punting to a followup, since we'd need to start writing `ext::fp` and whatnot everwhere.

There will likely be some work to sort out all the cargo features in a coherent fashion...

mcy (commented):
Sorry I took so long; meant to get to it yesterday but had exciting issues with my bike to take care of. Looks good to me overall.

README.md
mcy: Yup, agreed.

web_ext/rust_webserver/Cargo.toml
mcy: Does it really matter, though? The project is small enough that build times would not be particularly affected. I plan to do a pass over the subcrates to start documenting them, though, so maybe we can see if it makes sense to re-export some things or not as submodules of `ext` after I do that.

mcy: Got it. I'm happy to investigate a way to do improve the interface in the future. Like I said, I want to focus on writing docs first.

PR #49, opened 2021-10-24
Add some nominal types for representing algebra elements

mcy:
This is a WIP, since I expect it to tear up a lot of function signatures in `algebra`, `sseq`, and `ext`. I'm pushing it now to get feedback on the design.


mcy:
Just an update; I've been *super* hosed the last weeks and have been punting on this since I haven't been approaching this a great way. I think I'll take a look at redoing this by adding a separate, parallel `product` trait function that uses the new element types (and is just a thin wrapper) and gradually switch everything over to using those.

PR #52, opened 2021-11-25
Save format 2.0

### LINKED ISSUE #44: Revisit the saving mechanism
As we push into higher stems, I think it might be useful to rethink the way we store resolutions. Right now the `Resolution` struct is written to a file as-is. While that has its advantages, I can think of a few downsides.

- Very sensitive to the data layout. Any minor change is susceptible to make all prior archives unreadable, unless we're careful.
- Impossible to inspect manually.
- No support for incremental saving. This is the big one. When we start testing multi-day/week runs it will be inevitable that some crashes will happen, and it would be great to not lose the quasi-inverses that we will have spent many CPU-hours calculating. I think it would help to fragment the archive into smaller files. Maybe quasi-inverses could be stored in their own files?
- No compression. As the matrices get bigger the storage space becomes an issue. I've calculated that computing into the 200+ stems will require on the order of 10TB of disk space at the very least. It would be helpful to have the ability of writing/reading compressed data, especially since the archives are empirically very compressible. (I just checked and the resolution of S_2 up to t = 100 can be compressed by 60% using gzip.)

I don't have a concrete suggestion for a replacement right now, but I think it's something we should consider before pushing computations further (see #30).

dalcde:
This is a first draft that implements the ideas in
https://github.com/SpectralSequences/sseq/issues/44#issuecomment-978732350.

Currently, compression is not supported. I think the right way to do
this is that we don't compress when writing, but be able to read
compressed files.

The current implementation loads save files lazily. When we call
resolve_through_stem, for each bidegree, we check if we have data
available. If it is available, we simply load it. Otherwise, we compute
it from scratch as usual.

The save format is designed with the possibility of just mmap'ing the
FpVector limbs. I think mmap'ing a whole matrix is too much complication
as we have do deal with the indirection. rkyv can help you do this but I
don't think it is worth the effort.

A brief documentation of the file format is in ext/SAVE-FORMAT.md

One thing I noticed is that in general the filesystem has overhead for
having a file at all, even if it is tiny. For low degrees this save
format is actually highly inefficient.

Fixes #44


dalcde:
PS. The plan is to phase out the `saveload` module, since with the new formats we would want to do something with finer control.

Currently the secondary Steenrod alegbra code still uses `saveload`. I have a WIP complete rewrite of the secondary Steenrod algebra business, so I'll completely remove `saveload` when that happens.

Of course, we will also want to keep the saveload stuff for the migration script.

JoeyBF:
> Currently, compression is not supported. I think the right way to do
> this is that we don't compress when writing, but be able to read
> compressed files.

I agree, as long as we're able to compress what's already been written while the resolution is still being computed. That way we can have some kind of script that watches the output directory and compresses as it goes.

> I think mmap'ing a whole matrix is too much complication
> as we have do deal with the indirection. rkyv can help you do this but I
> don't think it is worth the effort.

I disagree. The reason I've been reworking `fp` is exactly because it's impossible to apply concurrent algorithms to a `Vec<FpVector>`. The "right" way is to have all rows in one big contiguous region, so that
  1. We save ourselves one level of indirection.
  2. The borrow checker allows us to have disjoint mutable borrows. Otherwise, when splitting a matrix in two vertical strips we would have to store a pair of `Vec<SliceMut<'_>>` and promise that they don't alias. That's impossible, at least with the current implementation of `SliceMut`.

So since matrices should be seen not as collections of vectors but as their own thing, they would need a separate deserialization scheme.

dalcde:
If you do matrices that way then it makes more sense to directly mmap it. The save file format I'm currently using would support that; all the FpVector limbs are contiguous.

Note that there are various places in the code where we assume it is cheap to add extra rows/columns. We might have to reorganize some of the exist code if we use a contiguous block. That said, I agree a contiguous block is the "right" way to do matrices.

> as long as we're able to compress what's already been written while the resolution is still being computed.
Apart from the kernels, once we write something we don't touch it ever again. Of course if we go with the `mmap` approach we have to be more careful. Personally I would probably just compress everything once I'm done with a calculation.

dalcde:
One feature implemented by this PR is that we can tell the resolution to not keep the quasi-inverses, but still compute them. If we throw away all the quasi-inverses after we save them, we can significantly improve the memory situation (and perhaps use mmap to read them in).

dalcde:
I'm pretty happy with the current state of the PR. I'll write a conversion
script and then merge the PR some time later today, if there are no further
comments.

Many existing scripts can be modified to make better use of this system, but I
think it makes sense to fix them whenever we actually want to use them.

dalcde:
I think one final thing to add is that we need a way to ensure we don't accidentally use a milnor save file with the adem basis or vice versa.

--

The conversion code is at https://github.com/SpectralSequences/sseq/tree/convert-save2.0 in the `convert` example. I will probably squash and force push this branch once the PR is merged.

For reference, using my resolution of S_2@milnor up to the 140th stem, the original data is ~40GB and the `zstd -22` compressed data is ~18GB, with a compression ratio of 44%. The compression took around 10 minutes utilizing 38 cores. The space usage breakdown is as follows:

```
95116      ./differentials
952        ./kernels
17990756   ./res_qis
93416      ./augmentation_qis
18180244   .
```
A lot of `differentials` and `augmentation_qis` disk usage is pure filesystem overhead. For example, `tar`'ing `differentials` results in a 26 MB archive.

Loading this (compressed) resolution used 1.9 GB and 67 seconds.

JoeyBF:
Are you saying that decompressing this 18GB archive took 67 seconds? Because if so that's even better than the raw read speed of the storage I'm using.

dalcde:
Nah I'm not loading the quasi-inverses. Should have mentioned that. (and 1.9 GB is the memory usage, which also gives that away)

JoeyBF:
Ah I see. How much time does it take to load the 40GB's worth of quasi-inverses from compressed storage?

dalcde:
I have not attempted, and I don't see myself doing that in the future. Most purposes can be satisfied by loading them on demand

JoeyBF:
I had the computation of the product structure in mind. Then I feel we would spend a significant amount of time loading and unloading quasi-inverses, especially if we make the process massively parallel. I guess we could solve that by some synchronization between the threads, to make sure they all use the same quasi-inverse at the same time.

dalcde:
I'd imagine this would require more careful architecture in any case, if we want to do this in a distributed way. We can easily run out of memory if we go past e.g. 150.

dalcde:
One issue with the current format is that in most filesystems, any file occupies at least 4kB of space, which is a huge overhead when storing small data (e.g. augmentation map quasi-inverses). If we want to store the results of many chain maps (e.g. to compute products) this could be even more problematic.

JoeyBF:
Right now I'm computing the product structure up to t = 190 using @jassch 's code that I parallelized using rayon. It's not distributed, only multi-threaded, but the memory overhead is insignificant compared to the amount needed to load the full resolution in the first place. The output files are mostly in the 10-100 kB range, so that could be a good way of storing the results. I think it would be worthwhile to include Jason in the discussion.

dalcde:
Do you store any of the intermediate chain maps? I'm thinking that saving each chain map of each bidegree would have a huge fs overhead.

JoeyBF:
Jason's original code saved the `ResolutionHomomorphism` objects to make resuming easier, but I deleted that part. It might be a good idea to store the lifts as `ChainMap`s, and it will also make it easier to reuse those when computing Massey products down the line. Right now the only output is the multiplication table, forgetting the actual lifts.

dalcde:
My original intention was to put the additional computed data (e.g. the resolution homomorphisms) in other subdirectories of the save directory in a similar format. This would make a lot of sense for e.g. the d2 data, but becomes really clumsy when we want to compute products. While "normal" products are reasonably fast, I think the Mod_{Cτ²} products will be quite a bit slower and the ability to resume would probably be quite useful.

dalcde:
P.S. it would be nice to get your modifications merged into the main tree. Your scripts would fit very well in the examples directory and CI can ensure future changes don't break them.

JoeyBF:
My guess is that we would have to try it to see if this is a big problem. Maybe grouping the small / all lifts in a tarball would help, at the cost of not being able to mmap anymore. IIRC there is a way of reading from a tar archive without extracting everything.

About the products, I think Jason would be the best person to ask. Right now the product computation is not really something that fits in a single file, and he has some useful code that it would make sense to integrate into the main library.

dalcde:
Another option is to use sqlite for the "small" things (i.e. differentials) and raw files for the large ones (i.e. quasi-inverses). We can still use the same serialization format; just using sqlite as our "index" instead of the filesystem.

dalcde:
The `rusqlite::blob` interface can potentially make this interact well with our existing code structure. We can then choose different storage backends for different kinds of data. For external applications all we need to know is that `open_save_path` and `create_save_path` give us something we can read from/write to. The main difference is that we need to supply a size to `create_save_path`.

JoeyBF:
I think we should see if storing lots of small files is as bad as we think, and if that's the case then sqlite sounds like it can get the job done for differentials. I still would much rather have quasi-inverses directly as separate files instead of storing them in that database.

dalcde:
I definitely agree with the quasi-inverses part. I think having a lot of
files for the differentials is fine, but having a lot of files for other
stuff is not (e.g. product chain maps for every single indecomposable).

dalcde:
One potential issue with sqlite is that different *processes* cannot write to the same database at the same time. Different threads is fine, which should satisfy many but not all use cases. If we use individual files, then as long as we are writing to different files we are good.

JoeyBF:
Ah that's a good point. If we start going distributed we might have to worry about that. My first thought would be some kind of dispatcher process that would interface between the file and other processes.

dalcde:
At a large enough scale it might be useful to use something like postgres.

dalcde:
The script at https://github.com/SpectralSequences/sseq/tree/convert-save2.0
now converts to the new format.

PR #53, opened 2021-11-26
Yet another vector rewrite

JoeyBF:
Here's a reworking of how we handle vectors, to set the stage for matrices later. I abstracted away the vector-like behavior that  `FpVector`, `Slice` and `SliceMut` shared and I put it in traits, `BaseVector` for immutable methods and `BaseVectorMut` for mutable ones. That lets us introduce some more vector-like structs down the line, like `Row` for a row of a matrix or `ArchivedFpVector` if we start using rkyv. I'm opening this as a draft because:

1. The ext library works but the rest of the repo has to be fixed still (mostly just importing those two traits).
2. It's still missing almost all the documentation.
3. There's a non-trivial performance hit when using this new module, and I would like to find why and fix it if possible.


dalcde:
Do you have concrete numbers RE the performance hit?

JoeyBF:
> 1. The ext library works

Actually I take that back. The ext library compiles, and all `fp` and `algebra` tests pass, but the resolution starts getting messed up at stem 19. So much so that the computation up to t = 40 already basically takes forever. The `fp` bench results are sane though, I can still share those if you want.

dalcde:
By "messed up" do you mean you get wrong results? In my experience that tends to make later calculations compute something ridiculous and take forever

JoeyBF:
Yeah, the rank of each bidegree explodes. Here's a sample output
```
Module (default: S_2):
Load from save? (optional):
Max s (default: 15):
Max t (default: 30):
Save file (optional):
·
·
·
·
·                                     :
·                                   · : ·
·                                 ·   : · ⠿
·                             ·   ·   · · ⠿ *
·                     ·       · · ·   · · ⠿ * *
·                   · ·     · · · ·   · : ⡿ * * *
·                 ·   ·     · :   · · · ∴ ⣿ * * * *
·             ·   ·         · ·   · : · ∴ ⡿ * * * * *
·     ·       · · ·         · ·   · · · · ⁙ ⁘ ⣿ * * * *
·   · ·     · · ·           · · ·   ·         : : : : * *
· ·   ·       ·               ·
·
```

JoeyBF:
Update: I tracked the error down to the debug_assert at matrix_inner.rs:747. The first time the inside of the for loop is executed is at t = 24, s = 3, and that's where the assert fails and our problems begin. Now I just need to check why it fails, since my reimplementation *should* be semantically equivalent to the current one.

ext/crates/fp/src/limb.rs
```diff
@@ -11,6 +11,133 @@ pub(crate) struct LimbBitIndexPair {
...
+    pub(crate) limbs: usize,
+    /// The total number of entries in the range.
+    pub(crate) logical: usize,
+    /// The index of the first entry in the first limb in the range. This is a number between 0
+    /// (inclusive) and [`entries_per_limb_const::<P>()`] (exclusive).
+    pub(crate) offset: usize,
+    /// The total number of limbs in the range, from the other base limb to the last limb of this
+    /// range.
+    pub(crate) total_limbs: usize,
+}
```
dalcde, line 34: This struct is `Copy` and quite large, which might be a source of performance impact.

ext/crates/fp/src/vector/generic.rs
```diff
@@ -0,0 +1,434 @@
...
+#[derive(Debug, Hash, Eq, PartialEq, Clone)]
+pub struct FpVectorP<const P: u32> {
+    pub(crate) limbs: Vec<Limb>,
+    pub(crate) len: LimbLength<P>,
+}
+
+/// A SliceP is a slice of an FpVectorP. This immutably borrows the vector and implements Copy
+#[derive(Debug, Copy, Clone)]
+pub struct SliceP<'a, const P: u32> {
+    pub(crate) ptr: *const Limb,
```
dalcde, line 23: Why does this have to be a raw pointer?

ext/crates/fp/src/limb.rs
JoeyBF: That's a good point, I'll experiment with that.

ext/crates/fp/src/vector/generic.rs
JoeyBF: So right now `Slice` and `SliceMut` store a pointer to the beginning of the vector in the ambient `FpVector`, so I totally agree that we could store references like the current code does.

However, in the future we'll want to have multiple `SliceMut`s on the same `FpVector` (for tiling, for example). If they all point to the same location, calling `_limbs_mut()` will create multiple mutable slices that all alias each other, and that's instant UB even if we're careful about only accessing the values we care about.

At that point I would like to make it so that the pointer actually points only to the beginning of the region containing the entries we care about. That way, calling `_limbs_mut()` will actually create mutable slices that are next to each other in memory but don't overlap.

dalcde: Wouldn't it be safe to have `&mut [Limb]` in that case? The point is that you check it is non-aliasing when you construct the `SliceMut` itself, instead of every time you use the `SliceMut`.

dalcde:
Here is a thought --- how different is BaseVector from `Into<Slice>` in the old world?

dalcde:
The correct trait bounds are really `for<'a> &'a T: Into<Slice<'a>>` and `for<'a> &'a mut T: Into<SliceMut<'a>>`. I used them here: https://github.com/SpectralSequences/sseq/blob/874fb2d04b323152c317603a2fe17ca59e93a07d/ext/crates/fp/src/matrix/quasi_inverse.rs#L75-L83
with implementations https://github.com/SpectralSequences/sseq/blob/874fb2d04b323152c317603a2fe17ca59e93a07d/ext/crates/fp/src/vector.rs#L427-L455

ext/crates/fp/src/vector/generic.rs
JoeyBF: That's a good point. I was ~~stealing~~ using the code for the `rulinalg` crate, and it uses raw pointers. I'll try to replace them by slices and see if it goes wrong. I'm thinking Rust might complain that we can't store a reference to a slice that we just created ourselves. I think storing a slice directly makes the struct a DST, which is awful. Maybe a `Box<[Limb]>`?

JoeyBF:
> Here is a thought --- how different is BaseVector from `Into<Slice>` in the old world?

Ideally we would want something like `impl<T: BaseVector> From<T> for Slice` but that would imply a conflicting implementation of `From<Slice>` on `Slice`. I still think it's important to distinguish the two, at least semantically. For example, this allows `FpVector` to have much more optimized implementations of several methods.

dalcde:
On Sun, Dec 05, 2021 at 03:08:41AM -0800, Joey Beauvais-Feisthauer wrote:
>That's a good point. I was ~~stealing~~ using the code for the `rulinalg` crate, and it uses raw pointers. I'll try to replace them by slices and see if it goes wrong. I'm thinking Rust might complain that we can't store a reference to a slice that we just created ourselves. I think storing a slice directly makes the struct a DST, which is awful. Maybe a `Box<[Limb]>`?

Storing [Limb] is what makes it a DST. A &[Limb] is fine because it is a
pointer to a DST, and has a known size (namely 2 * 64 bits).

dalcde:
Why would there be a conflict? I would be surprised if the FpVector
improvement would be significant, especially if the matrix slices are
going to be slices anyway. (otoh, an AlignedSlice struct would be nice
and good for semantics).

For maintenance puprposes, I think there is definitely a benifit to
having a smaller code footprint if possible.

JoeyBF:
> Storing [Limb] is what makes it a DST. A &[Limb] is fine because it is a pointer to a DST, and has a known size (namely 2 * 64 bits).

Of course. I was thinking that `std::slice::from_raw_parts` returned a raw slice instead of a reference; since it's just a reference, there should be no issue storing it.

> Why would there be a conflict? I would be surprised if the FpVector improvement would be significant, especially if the matrix slices are going to be slices anyway. (otoh, an AlignedSlice struct would be nice and good for semantics). For maintenance puprposes, I think there is definitely a benifit to having a smaller code footprint if possible.

Any `T` implements `From<T>`, but since `Slice` also implements `BaseVector` that would give us a second implementation of `From<Slice>` on `Slice`. I haven't tried yanking out the special `FpVector` methods but my intuition would tell me that they give a decent speedup.

What did you have in mind for an `AlignedSlice` struct? Something like a struct that is guaranteed to begin/end on a limb boundary? Logically that would be what is returned by some `split_borrow` method on `FpVector`.

dalcde:
On Sun, Dec 05, 2021 at 10:36:22PM -0800, Joey Beauvais-Feisthauer wrote:
>> Why would there be a conflict? I would be surprised if the FpVector improvement would be significant, especially if the matrix slices are going to be slices anyway. (otoh, an AlignedSlice struct would be nice and good for semantics). For maintenance puprposes, I think there is definitely a benifit to having a smaller code footprint if possible.
>
>Any `T` implements `From<T>`, but since `Slice` also implements `BaseVector` that would give us a second implementation of `From<Slice>` on `Slice`. I haven't tried yanking out the special `FpVector` methods but my intuition would tell me that they give a decent speedup.

I was thinking about *defining* `BaseVector` to mean `Into<Slice>`.
Alternatively, `BaseVector` can auto-implement all other methods based
on the `as_slice` method, which leaves room for more optimized versions.

>What did you have in mind for an `AlignedSlice` struct? Something like a struct that is guaranteed to begin/end on a limb boundary? Logically that would be what is returned by some `split_borrow` method on `FpVector`.

Yeah

dalcde:
Just as a note, after #80 we cannot make e.g. Module::act take in `impl BaseVector` since it would no longer be object safe.

ext/crates/fp/src/vector/generic.rs
JoeyBF: Fixed by [7fe0f9f](https://github.com/SpectralSequences/sseq/pull/53/commits/7fe0f9ff77b3308a14b088145ade4eda9d367857)

ext/crates/fp/src/limb.rs
JoeyBF: Fixed in [986b6b6](https://github.com/SpectralSequences/sseq/pull/53/commits/986b6b6565f94eca0d172cd0aed87d29dbd59ca4) (mostly)

JoeyBF:
All tests pass except the selenium ones, and they don't look like the same flakey ones we used to have. I'm not sure how to fix them, since at first glance they don't seem related to `fp`

dalcde:
empirically, does the web interface work?

JoeyBF:
I'm not sure how I should check that locally. How do I start the webserver?

dalcde:
cargo run in the directory should do the job

JoeyBF:
It seems adding a differential called a code path in fp that wasn't tested, but it's fixed now

dalcde:
Do you mind rebasing and resolving merge conflicts first?

JoeyBF:
I can also squash all the commits if you would prefer less noise. These days I tend to prefer force-pushing as little as possible

JoeyBF:
Note that miri in the CI for 65d570c is complaining about `OnceBiVec::par_extend` and I have no idea why. I assume it's either a false positive, or the new nightly miri is picking up something that used to fly under the radar

dalcde:
I think squash + rebase would be good. The diff is currently quite messy

JoeyBF:
I squashed the commits and rebased on master. There were a few interesting tidbits in the commit messages so I put them in the code as comments instead

ext/crates/fp/src/prime.rs
```diff
@@ -506,11 +507,18 @@ impl Iterator for BinomialIterator {
...
 mod tests {
     use super::*;
+    #[cfg(feature = "odd-primes")]
     use crate::constants::PRIMES;
 
     #[test]
     fn inverse_test() {
-        for &p in PRIMES.iter() {
+        #[cfg(feature = "odd-primes")]
+        let primes = PRIMES;
```
dalcde, line 516: Can PRIMES be [2] when odd-primes is false? (And maybe similarly have a TEST_PRIMES that is either [2] or [2, 3, 5, 7, 11])

ext/crates/fp/src/vector/specialized.rs
```diff
@@ -0,0 +1,502 @@
+//! This module replaces `specialized` when `odd-primes` is disabled. Instead of producing enum
+//! wrappers, it simply rexports `FooP<2>` as `Foo`.
```
dalcde, line 2: I assume this is supposed to go to specialized_2.rs?

ext/crates/fp/src/limb.rs
```diff
@@ -97,7 +261,7 @@ pub(crate) fn reduce<const P: u32>(limb: Limb) -> Limb {
 
 /// Check whether or not a limb is reduced, i.e. whether every entry is a value in the range `0..P`.
 /// This is currently **not** faster than calling [`reduce`] directly.
-pub(crate) fn is_reduced<const P: u32>(limb: Limb) -> bool {
+pub(crate) fn _is_reduced<const P: u32>(limb: Limb) -> bool {
```
dalcde, line 264: What is the leading underscore for?

ext/crates/fp/src/limb.rs
```diff
@@ -1,7 +1,6 @@
 use std::ops::Range;
```
dalcde, line 1: I think it would be better to make everything in this module `pub(crate)`

ext/crates/fp/src/limb.rs
```diff
@@ -11,6 +10,148 @@ pub(crate) struct LimbBitIndexPair {
...
+        self.end -= offset * entries_per;
+        offset
+    }
+
+    pub fn restrict_to(&self, other: Self) -> Self {
+        debug_assert!(self.start + other.end <= self.end);
+        Self::from_start_end(other.start + self.start, other.end + self.start)
+    }
+
+    pub fn trim_start(&self, offset: usize) -> Self {
```
dalcde, line 94: The implementation is actually trim_start().apply_shift(). I would guess the compiler can optimize trm_start().apply_shift() to the current implementation (but please double check!); Would it be better to have trim_start do the obvious thing instead?

ext/crates/fp/src/vector/internal/mod.rs
```diff
@@ -0,0 +1,686 @@
...
+pub trait InternalBaseVectorP<const P: u32>: Sized {
+    /// Returns a pointer to the allocation containing the actual data. This is a raw pointer and
+    /// does not take lifetimes into account. It is the responsibility of the caller to ensure that
+    /// the pointer is not dereferenced after the allocation is freed.
+    ///
+    /// We use a pointer instead of a slice because otherwise handling the lifetimes is a huge mess.
+    /// In practice it is almost always better to use [`InternalBaseVectorP::_limbs`] to manipulate
+    /// the underlying data.
+    fn _as_ptr(&self) -> *const Limb;
+
```
dalcde, line 21: It seems like this is only used to implement _limbs and _slice, and _slice can be implemented in terms of _limbs. Is there really a need for this function?

ext/crates/fp/src/limb.rs
```diff
@@ -11,6 +10,148 @@ pub(crate) struct LimbBitIndexPair {
...
+
+    /// Shift the entire `LimbLength` backwards so that the start of the range belongs to the first
+    /// limb, and return the number of limbs shifted.
+    pub fn apply_shift(&mut self) -> usize {
+        let entries_per = entries_per_limb_const::<P>();
+        let offset = self.start / entries_per;
+        self.start -= offset * entries_per;
+        self.end -= offset * entries_per;
+        offset
+    }
```
dalcde, line 87: Would it be cleaner for this to return a new LimbLength and the offset as a tuple? No other functions here mutate the LimbLength

ext/crates/fp/src/limb.rs
```diff
@@ -11,6 +10,148 @@ pub(crate) struct LimbBitIndexPair {
     pub(crate) bit_index: usize,
 }
 
+/// A struct that defines a range of entries in a slice of limbs. We number limbs starting from some
+/// other limb that might not be in the range. This is because we must be able to index a struct
+/// using indices for another ambient struct. For example, accessing a `Slice` using coordinates
+/// from an ambient `FpVector`.
+#[derive(Debug, Hash, PartialEq, Eq, Clone, Copy)]
+pub struct LimbLength<const P: u32> {
```
dalcde, line 18: This is bikeshedding territory but I would say "LimbRange" is a better name.

ext/crates/fp/src/limb.rs
```diff
@@ -11,6 +10,148 @@ pub(crate) struct LimbBitIndexPair {
...
+
+    #[inline]
+    pub fn limb_range(&self) -> Range<usize> {
+        range::<P>(self.start, self.end)
+    }
+
+    /// This function underflows if `self.start + self.logical() == 0`, which happens if and only if
+    /// we are taking a slice of width 0 at the start of an `FpVector`. This should be a very rare
+    /// edge case. Dealing with the underflow properly would probably require using `saturating_sub`
+    /// or something of that nature, and that has a nontrivial (10%) performance hit.
```
dalcde, line 118: Can we insert a debug_assert that this doesn't happen? I assume it would be a programmer error if they allow this to happen. And the docs can simply say that the function cannot be called on such a range.

Fun fact: For a long time, `FpVector::<2>::add` was incorrectly implemented to ignore the coefficient, since we often check if the coefficient is 0 before we call `add`. It was discovered when it lead to a very confusing bug.

ext/crates/fp/src/limb.rs
```diff
@@ -11,6 +10,148 @@ pub(crate) struct LimbBitIndexPair {
...
+    /// The total number of limbs in the range.
+    ///
+    /// We store this value instead of computing it on the fly because benchmarks tend to show that
+    /// the tradeoff is beneficial in high dimensions (>1000). We might want to only enable this
+    /// when odd-primes is enabled, since computing this number is easier when `p == 2`, so the
+    /// tradeoff is potentially worse.
+    limbs: usize,
+}
+
+impl<const P: u32> LimbLength<P> {
```
dalcde, line 34: It seems like a lot of the functions can be made `const`, e.g. `bit_offset`. It's probably good to make them so.

ext/crates/fp/src/limb.rs
```diff
@@ -11,6 +10,148 @@ pub(crate) struct LimbBitIndexPair {
...
+        Self {
+            start: 0,
+            end: logical,
+            limbs,
+        }
+    }
+
+    /// Returns a `LimbLength` describing a vector starting at entry `start` and ending at entry
+    /// `end`. Most methods assume that `self.start < entries_per_limb_const::<P>()`, so it advised
+    /// to call `apply_shift` before using the return value.
```
dalcde, line 55: What exactly are these "most methods"? It seems like e.g. `bit_offset` makes such an assumption. I think this should be in the documentation of the functions that make these assumptions rather than `from_start_end`.

ext/crates/fp/src/limb.rs
```diff
@@ -46,6 +187,29 @@ pub(crate) const fn entries_per_limb_const<const P: u32>() -> usize {
     BITS_PER_LIMB / bit_length_const::<P>()
 }
 
+/// This is identical to [`limb::number`], except that it's not const. Hopefully almost every method
+/// in the limb crate can be const once the matrix rewrite is in place.
```
dalcde, line 191: We can probably implement the const generics version in terms of this. This can probably be a const fn as well.

ext/crates/fp/src/vector/internal/mod.rs
```diff
@@ -0,0 +1,686 @@
...
+        if limb_range.is_empty() {
+            return true;
+        }
+        let (min_mask, max_mask) = self._len().limb_masks();
+        if self._limbs()[limb_range.start] & min_mask != 0 {
+            return false;
+        }
+
+        let inner_range = self._len().limb_range_inner();
+        if !inner_range.is_empty() && self._limbs()[inner_range].iter().any(|&x| x != 0) {
```
dalcde, line 49: Is the first check necessary/helpful?

ext/crates/fp/src/vector/mod.rs
```diff
@@ -1,491 +1,39 @@
...
+pub mod generic;
+pub(crate) mod internal;
+
+#[cfg(feature = "odd-primes")]
+pub mod specialized;
+pub mod specialized_2;
+
+pub use specialized::{FpVector, FpVectorNonZeroIterator, Slice, SliceMut};
+#[cfg(not(feature = "odd-primes"))]
+pub use specialized_2 as specialized;
```
dalcde, line 11: I think you want to swap the order of these two statements.

dalcde (commented):
How does this rewrite affect performance?

ext/crates/fp/src/limb.rs
JoeyBF: It's only used in `_truncate`, which is itself never used. Without the underscore clippy complains about dead code. I believe it would be used in `BaseVectorMut::add_truncate`, but that one is unimplemented right now. I should probably fix that if we want to use `algebra::PolynomialAlgebra::multiply_monomials` in the future.

ext/crates/fp/src/limb.rs
JoeyBF: Everything can be, and should, except for `LimbLength` itself. It's technically in the API, since (e.g.) `BaseVectorP` extends `InternalBaseVectorP` which itself contains a method returning a `LimbLength`.

ext/crates/fp/src/vector/internal/mod.rs
JoeyBF: I think it is, otherwise `self._limbs()[inner_range]` panics IIRC

ext/crates/fp/src/vector/mod.rs
JoeyBF: Wow, yeah I don't know how I didn't catch that before

ext/crates/fp/src/limb.rs
JoeyBF: Good idea, I replaced the code for `number`

ext/crates/fp/src/limb.rs
JoeyBF: I was a bit lazy to check if the assumption was needed in each case, since it's only an internal struct, but it should be good now

ext/crates/fp/src/limb.rs
JoeyBF: I made a few more functions `const`. The others can't be because either they end up calling `pack` which has a for loop, or they call some non-const function in stdlib

ext/crates/fp/src/limb.rs
JoeyBF: I made the documentation hopefully a bit more explicit

ext/crates/fp/src/limb.rs
JoeyBF: I went through a few iterations with the naming. I initially went with `ColumnSize` since we'll be reusing that struct to define the columns of our matrices, but `ColumnRange`, `LimbRange`, etc. are all good options I think

ext/crates/fp/src/limb.rs
JoeyBF: Agreed. It technically causes one more unnecessary copy but the compiler should catch that, and in any case it's not performance critical

ext/crates/fp/src/vector/specialized.rs
JoeyBF: Thanks!

ext/crates/fp/src/prime.rs
JoeyBF: Actually I realized that most problems are fixed if we just let `num_primes = 1` in build.rs when odd-primes is disabled. Thanks for pointing that out

ext/crates/fp/src/limb.rs
JoeyBF: That might complicate things slightly if we only assume `self.start % entries_per_limb_const::<P>() == 0`, e.g. depending on which of `self.start` or `offset` is bigger. I supposed that can be solved by a `saturating_sub`.

The reason I would tend to leave it that way is because we only use it when trimming the front of an `FpVector`, where the start is 0 by definition

ext/crates/fp/src/vector/internal/mod.rs
JoeyBF: That kind of mirrors the `InternalBaseMatrixP` trait that does need a raw pointer. I figure already having a raw pointer here might make the interplay easier.

I tried just now to yank `_as_ptr` and make everything depend on `_limbs` but I'm running into a bunch of issues due to lifetimes. There might be a way of handling that gracefully that I'm not seeing, probably because it's getting late

ext/crates/fp/src/vector/internal/mod.rs
dalcde: This seems to work fine: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&code=fn%20main()%20%7B%0A%20%20%20%20let%20x%20%3D%20%5B1%2C%202%2C%203%5D%3B%0A%20%20%20%20println!(%22%7B%3A%3F%7D%22%2C%20%26x%5B0..0%5D)%0A%7D

ext/crates/fp/src/limb.rs
dalcde: I think it is better to use `#[allow(dead_code)]` and document why it is dead

ext/crates/fp/src/vector/internal/mod.rs
dalcde: What lifetime issues are you running into?

ext/crates/fp/src/limb.rs
```diff
@@ -11,6 +10,146 @@ pub(crate) struct LimbBitIndexPair {
...
+    /// # Panics
+    ///
+    /// This function underflows if `self.start + self.logical() == 0`, which happens if and only if
+    /// we are taking a slice of width 0 at the start of a limb. This should be a very rare edge
+    /// case. Dealing with the underflow properly would probably require using `saturating_sub` or
+    /// something of that nature, and that has a nontrivial (10%) performance hit.
+    #[inline]
+    pub(crate) fn limb_range_inner(&self) -> Range<usize> {
+        let range = self.limb_range();
+        debug_assert!(range.end > 0, "Underflow");
```
dalcde, line 117: I think this can have a more descriptive error message (or none at all; the current one isn't very helpful)

ext/crates/fp/src/limb.rs
dalcde: I think these kind of unspoken assumptions can be very dangerous, especially if other people modify the code

ext/crates/fp/src/vector/internal/mod.rs
JoeyBF: You're right, originally it was guarding against ranges like `1..0`, but those only happen if `limb_range` is already empty, which we now check beforehand.

ext/crates/fp/src/vector/internal/mod.rs
JoeyBF: So I just made the changes and it looks like `fp` works just fine, but the functionality is slightly different. I was actually running into that same issue before I realized I needed that `Self: 'a` bound. The `algebra::free_module::MuFreeModule::slice_vector` method returns a slice of a slice, and the compiler complains that `v` doesn't live long enough. Having that `Self: 'a` bound makes it behave more like a real `std::slice` slice.

ext/crates/fp/src/limb.rs
JoeyBF: I think I added the documentation in the right places now

ext/crates/fp/src/limb.rs
JoeyBF: Done

ext/crates/fp/src/limb.rs
JoeyBF: Actually I just remembered that overflows and underflows are always checked in debug mode anyway. So we can remove that

ext/crates/fp/src/limb.rs
dalcde: I think the "it [is] advised to call `apply_shift` before using the return value" can now be removed

ext/crates/fp/src/vector/internal/mod.rs
```diff
@@ -671,13 +671,15 @@ pub trait InternalBaseVectorMutP<const P: u32>: InternalBaseVectorP<P> {
...
-
+    /// This method is only implemented on `FpVectorP` right now. This is the only use case so far,
+    /// so I don't feel too bad about marking them as unimplemented.
     fn _copy_from_slice(&mut self, _slice: &[u32]) {
         unimplemented!();
     }
 
+    /// This function is currently only used in [`algebra::PolynomialAlgebra::multiply_monomials`],
+    /// on [`FpVectorP`]. We mostly care about the Steenrod algebra, which doesn't use this code, so
+    /// I'm marking it as unimplemented for now.
```
dalcde, line 682: @hoodmane should confirm but my understanding is that PolynomialAlgebra is useful for implementing Steenrod modules that are polynomial algebras (e.g. cohomology of BU(n)).

ext/crates/fp/src/limb.rs
dalcde: Do you want to add that to the documentation?

ext/crates/fp/src/vector/internal/mod.rs
dalcde: I'm not convinced InternalBaseMatrixP requires a raw pointer. In any case I think we should minimize the amount of unsafe needed.

dalcde: (the pointer can be obtained from _limbs anyway)

ext/crates/fp/src/vector/internal/mod.rs
hoodmane: Yeah. This is in particular used for divided power algebras and exterior algebras. The name `PolynomialAlgebra` really means a tensor of divided power, exterior, and polynomial parts. It was intended for these to also have streenrod algebra structure. Once upon a time I had the Dixon algebra because of a special request, but doing BU(n) and U(n) is more reasonable. It's worth noting though that I never quite got the sign rule correct at odd primes.

Also it looks like the definition of `add_truncate` isn't even currently correct, probably because the name is cryptic and it has no test coverage.

ext/crates/fp/src/vector/internal/mod.rs
JoeyBF: It needs to because in the case of a `MatrixSlice` the limbs won't form a contiguous region in memory if it has fewer columns than the ambient matrix. I also would rather avoid unsafe code but I don't see how we can have the appropriate lifetimes that make `slice_vector` compile.

ext/crates/fp/src/limb.rs
JoeyBF: Done

ext/crates/fp/src/limb.rs
JoeyBF: Done

JoeyBF:
> How does this rewrite affect performance?

I'm still running benchmarks but I'll comment on that shortly

ext/crates/fp/src/vector/internal/mod.rs
JoeyBF: Does that mean that I should implement `add_truncate` anyway? I'm not concerned with other modules than F_2 personally but I wouldn't want to break the code for someone who uses ext to compute BU(n)

hoodmane: Since it's already broken it doesn't seem to me that this PR is making the situation any worse.

JoeyBF: Ah, got it. I thought it was only broken for odd primes. I could even remove the dead code from the limbs module then

hoodmane: Actually I take it back it works correctly I was just reading the code wrong.

JoeyBF: Alright, I added the implementation when the source is an `FpVector` and the target starts on a limb boundary

JoeyBF:
Here are the benchmarks, compared using critcmp
```
group                             master                                 new_vector
-----                             ------                                 ----------
row_reduce_2/row_reduce_2_10      1.00   565.5±27.64ns        ? ?/sec    1.11   627.9±33.43ns        ? ?/sec
row_reduce_2/row_reduce_2_100     1.00     15.9±0.67µs        ? ?/sec    1.03     16.3±1.35µs        ? ?/sec
row_reduce_2/row_reduce_2_1000    1.00  1163.2±31.80µs        ? ?/sec    1.05  1224.1±76.33µs        ? ?/sec
row_reduce_2/row_reduce_2_20      1.00  1334.2±45.95ns        ? ?/sec    1.26  1674.6±272.55ns        ? ?/sec
row_reduce_2/row_reduce_2_2000    1.00      5.0±0.13ms        ? ?/sec    1.05      5.2±0.13ms        ? ?/sec
row_reduce_2/row_reduce_2_4000    1.00     29.5±0.93ms        ? ?/sec    1.02     29.9±1.57ms        ? ?/sec
row_reduce_2/row_reduce_2_420     1.00    205.2±5.12µs        ? ?/sec    1.17   241.1±47.42µs        ? ?/sec
row_reduce_2/row_reduce_2_69      1.00      8.6±0.23µs        ? ?/sec    1.09      9.3±1.15µs        ? ?/sec
row_reduce_3/row_reduce_3_10      1.00   971.7±41.19ns        ? ?/sec    1.35  1313.6±52.65ns        ? ?/sec
row_reduce_3/row_reduce_3_100     1.00    110.5±3.71µs        ? ?/sec    1.38    152.5±1.55µs        ? ?/sec
row_reduce_3/row_reduce_3_1000    1.13     35.2±0.93ms        ? ?/sec    1.00     31.1±1.02ms        ? ?/sec
row_reduce_3/row_reduce_3_20      1.00      3.5±0.10µs        ? ?/sec    1.46      5.1±0.18µs        ? ?/sec
row_reduce_3/row_reduce_3_420     1.00      3.4±0.07ms        ? ?/sec    1.12      3.8±0.43ms        ? ?/sec
row_reduce_3/row_reduce_3_69      1.00     48.1±1.12µs        ? ?/sec    1.45     69.8±1.76µs        ? ?/sec
row_reduce_5/row_reduce_5_10      1.00   990.6±26.66ns        ? ?/sec    1.22  1213.4±61.51ns        ? ?/sec
row_reduce_5/row_reduce_5_100     1.00    153.9±3.17µs        ? ?/sec    1.15   177.2±24.86µs        ? ?/sec
row_reduce_5/row_reduce_5_1000    1.30     65.6±2.12ms        ? ?/sec    1.00     50.6±1.57ms        ? ?/sec
row_reduce_5/row_reduce_5_20      1.00      3.8±0.04µs        ? ?/sec    1.32      5.0±0.20µs        ? ?/sec
row_reduce_5/row_reduce_5_420     1.18      6.2±0.05ms        ? ?/sec    1.00      5.3±0.38ms        ? ?/sec
row_reduce_5/row_reduce_5_69      1.00     59.9±1.93µs        ? ?/sec    1.22     72.8±2.89µs        ? ?/sec
row_reduce_7/row_reduce_7_10      1.00  1810.6±76.17ns        ? ?/sec    1.19      2.2±0.05µs        ? ?/sec
row_reduce_7/row_reduce_7_100     1.00   486.1±49.60µs        ? ?/sec    1.12   542.8±70.90µs        ? ?/sec
row_reduce_7/row_reduce_7_1000    1.07   318.7±10.00ms        ? ?/sec    1.00   296.8±14.54ms        ? ?/sec
row_reduce_7/row_reduce_7_20      1.00      9.1±0.15µs        ? ?/sec    1.23     11.2±0.16µs        ? ?/sec
row_reduce_7/row_reduce_7_420     1.06     25.4±1.29ms        ? ?/sec    1.00     23.9±0.58ms        ? ?/sec
row_reduce_7/row_reduce_7_69      1.00    177.7±4.62µs        ? ?/sec    1.13    201.0±6.65µs        ? ?/sec
```

It looks like it tends to be better for bigger matrices, but worse otherwise. Also not great for p = 2. I was expecting something roughly like this, once new matrices will be in place the performance should go back up with interest. I didn't bench all of ext however.

dalcde:
Do you know/have an idea where the new overhead comes from?

JoeyBF:
I'm trying to run iai to figure it out but cachegrind keeps crashing. Could you confirm that you also get `Failed to run benchmark in cachegrind. Exit code: signal: 4` when running `cargo bench --bench iai` in the ext folder on the master branch?

JoeyBF:
Actually, even compiling `cargo build --release --example resolve` and trying `valgrind --tool=cachegrind target/release/examples/resolve` gives me an illegal instruction error

JoeyBF:
I figured it was a recent instruction set that was tripping up valgrind so I ran iai with target CPU x86-64. I'm having trouble reading the results, but it seems that cache locality is a bit worse and there are more instructions in general when the dimensions are small, but fewer than master when the dimensions get bigger. Here's [the comparison](https://github.com/SpectralSequences/sseq/files/8999022/iai_diff.txt); master on the left, this PR on the right. That's probably all I can say without diving into the asm

ext/crates/fp/src/vector/internal/mod.rs
dalcde: Do we have a need for slicing by columns? I remember this was desired so that we can shard row reduction, but it seems like this is no longer necessary with Nassau's algorithm?

dalcde: (or rather, do we have a need for holding multiple mutable slices in disjoint columns?)

JoeyBF: I think we definitely need that if we want concurrent algorithms. We probably don't need cutting-edge row reduction algorithms anymore but we can still implement fast matrix-matrix operations. I believe there's a few places where it can be useful, but I'm not familiar enough with the code outside of `fp`. One idea was to use exactly that for #102

dalcde:
I don't think that's the right direction to do concurrency in. Have you benchmarked how long each write_qi takes? My guess would be that each call is very fast, but we call it many many times.

JoeyBF:
What would you mean by very fast? Looking at top I would guess something like 30s-1m maybe. That's also a big part of why I think #101 would be useful

JoeyBF:
Closing as superseded. Every piece of this landed by another route:

- the `limb` module rework → `fp/src/limb.rs`
- `add_truncate` → `fp/src/vector/fp_wrapper/` and `impl_fqvector.rs`
- SIMD in `add_shift_none` → `impl_fqslicemut.rs`, plus a dedicated `fp/src/simd`
- macro generation via build script → `fp/build.rs`
- removing the hardcoded max prime of 7 → no such hardcoding remains

`LimbLength` is the one name that didn't survive, but `fp/src/` has since been reorganised into `field/`, `vector/`, `prime/`, `matrix/` — a layout that didn't exist when this branch was written. Rebasing 70 files across that restructure would cost more than redoing anything still wanted. The remote branch stays if anyone wants to read it.

PR #54, opened 2021-11-28
Use the Milnor basis by default

JoeyBF:
As we discussed, it shouldn't break the saves for people who use the webserver, and the Milnor basis is just better in general.


dalcde:
I think you need to set `"algebra" = ["adem"]` on the `A-mod-*.json` modules.

JoeyBF:
Hmm, it looks like the tests fail at the same place even though I added that line. Do you think I need to add it to the other modules too?

JoeyBF:
Thanks Dexter for working through all the tests! I wouldn't have expected this change to break so many things, but I suppose that means that the tests were really doing their job.

PR #68, opened 2022-02-01
Run formatters on chart code

hoodmane:
It would be good to add format checks to CI.


hoodmane:
For some reason my .gitignore doesn't seem to be working as expected...

PR #70, opened 2022-02-11
Algebra selection algorithm

JoeyBF:
I implemented some algorithm to choose subalgebras. It might still be suboptimal, and I haven't mathematically checked that it's valid but empirically it produces the right data.

### DIFF
--- a/ext/src/nassau.rs
+++ b/ext/src/nassau.rs
@@ -139,16 +139,16 @@ impl MilnorSubalgebra {
     fn top_degree(&self) -> i32 {
         self.profile
             .iter()
-            .map(|&entry| 2i32.pow(entry as u32) - 1)
+            .map(|&entry| (2 << entry) - 1)
             .enumerate()
-            .map(|(idx, entry)| (2i32.pow(idx as u32 + 1) - 1) * entry)
+            .map(|(idx, entry)| ((2 << (idx + 1)) - 1) * entry)
             .sum()
     }
 
     fn optimal_for(s: u32, t: i32) -> MilnorSubalgebra {
         let mut result = MilnorSubalgebra::zero_algebra();
         for subalgebra in SubalgebraIterator::new() {
-            let coeff = 2i32.pow(subalgebra.profile.len() as u32) - 1;
+            let coeff = (2 << subalgebra.profile.len()) - 1;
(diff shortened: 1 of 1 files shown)


ext/src/nassau.rs
```diff
@@ -130,6 +135,67 @@ impl MilnorSubalgebra {
...
+        let mut subalgebras = vec![MilnorSubalgebra::zero_algebra()];
+        for subalgebra in SubalgebraIterator::new() {
+            let coeff = 2i32.pow(subalgebra.profile.len() as u32) - 1;
+            if t < coeff * (s as i32 + 1) + subalgebra.top_degree() {
+                // (s,t) is not in the vanishing region of `subalgebra` or any further subalgebra
+                break;
+            }
+            subalgebras.push(subalgebra);
+        }
+        subalgebras.pop().unwrap()
```
hoodmane, line 158: I don't understand why you have a `subalgebras` vector if we only care about the last element. Wouldn't the following make more sense:
```suggestion
        let mut result = MilnorSubalgebra::zero_algebra();
        for subalgebra in SubalgebraIterator::new() {
            let coeff = 2i32.pow(subalgebra.profile.len() as u32) - 1;
            if t < coeff * (s as i32 + 1) + subalgebra.top_degree() {
                // (s,t) is not in the vanishing region of `subalgebra` or any further subalgebra
                break;
            }
            result = subalgebra;
        }
        result
```

ext/src/nassau.rs
```diff
@@ -130,6 +135,67 @@ impl MilnorSubalgebra {
     pub fn iter_signatures(&self, degree: i32) -> impl Iterator<Item = Vec<PPartEntry>> + '_ {
         SignatureIterator::new(self, degree)
     }
+
+    fn top_degree(&self) -> i32 {
+        self.profile
+            .iter()
+            .map(|&entry| 2i32.pow(entry as u32) - 1)
```
dalcde, line 142: I personally prefer using bitshift operators, which is what most of the codebase does. I don't have very strong opinions though.

JoeyBF:
I'll add also that this is suboptimal in the sense that we only work below the vanishing line. In theory we could also work above it, I don't know how much of a speedup that would give.

PR #72, opened 2022-02-13
Add logging and timings to Nassau's algorithm

JoeyBF:
I also added a `Display` implementation for `MilnorSubalgebra`.

### DIFF
--- a/ext/src/nassau.rs
+++ b/ext/src/nassau.rs
@@ -163,20 +163,13 @@ impl MilnorSubalgebra {
 
 impl Display for MilnorSubalgebra {
     fn fmt(&self, out: &mut std::fmt::Formatter<'_>) -> std::result::Result<(), std::fmt::Error> {
-        let name = if let Ok(a_name) = match self.profile[..] {
-            [] => Ok("F_2"),
-            [1] => Ok("A(0)"),
-            [2, 1] => Ok("A(1)"),
-            [3, 2, 1] => Ok("A(2)"),
-            [4, 3, 2, 1] => Ok("A(3)"),
-            [5, 4, 3, 2, 1] => Ok("A(4)"),
-            _ => Err(()),
-        } {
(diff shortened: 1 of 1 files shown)


ext/src/nassau.rs
```diff
@@ -159,6 +161,25 @@ impl MilnorSubalgebra {
...
+            [3, 2, 1] => Ok("A(2)"),
+            [4, 3, 2, 1] => Ok("A(3)"),
+            [5, 4, 3, 2, 1] => Ok("A(4)"),
+            _ => Err(()),
+        } {
+            a_name.to_string()
+        } else {
+            format!("Algebra with profile {:?}", self.profile)
+        };
+        write!(out, "{}", name)
```
dalcde, line 179:
```suggestion
        if self.profile.is_empty() {
            write!(out, "F_2")
        } else if self.profile.len() == self.profile[0] {
            write!(out, "A({})", self.profile.len())
        } else {
            write!(out, "Algebra with profile {self.profile:?}")
        }
```
Something like this should work (untested)

ext/src/nassau.rs
```diff
@@ -440,6 +462,12 @@ impl Resolution {
             assert!(dx.is_zero(), "dx non-zero at t = {t}, s = {s}");
         }
         self.differential(s).add_generators_from_rows(t, xs);
+        let stop_the_clock = timer.elapsed();
+        eprintln!(
+            "[{:>6}.{:>06} s] Computed bidegree ({s}, {t}) with {subalgebra}",
+            stop_the_clock.as_secs(),
+            stop_the_clock.subsec_micros()
+        );
```
dalcde, line 470: Don't have to change it in this PR, but I usually like to print the data in csv format for further postpocessing, e.g. generating graphs

dalcde:
I'm happy with either.

PR #78, opened 2022-03-06
Fix python-ext

hoodmane:
So far this gets the algebra crate compiling successfully. Not sure whether it works or not.

TODO:

- [ ] resolution
- [ ] resolution-homomorphism
- [ ] upgrade Pyo3 to 0.16.1


dalcde:
I think it would be helpful to get partially wokring code merged, together with
some tests to prevent breakage, to avoid merge conflicts piling up.

PR #80, opened 2022-03-08
Replace FiniteModule with dynamic dispatch

dalcde:
This cleans up quite a bit of boilerplate code.

This makes resolve_through_stem slightly slower (~2-4%) without nassau, and has no performance impact with nassau. Since large computations should be done with nassau, and the impact is probably asymptotically negligible, I think this is fine.

@hoodmane do you mind having a look and see if this will be problematic for the python bindings? (I hope it makes things better, but you never know...)


dalcde:
Merged manually

PR #88, opened 2022-03-28
Delete open files when ctrl-c received

### LINKED ISSUE #79: Implement signal handling
I think it would be a good idea to add some handlers to make sure we don't leave a save directory in a dirty state. I'm thinking of intercepting:

 - SIGINT/SIGTERM to remove any malformed files before exiting.
 - SIGUSR1 to dump a summary of ongoing computations to stderr (if `logging` is enabled).

Maybe we could also intercept SIGSTP to print the same data as SIGUSR1.

dalcde:
Partially fixes #79 .

@JoeyBF comments?

### DIFF
--- a/ext/Cargo.toml
+++ b/ext/Cargo.toml
@@ -28,6 +28,9 @@ byteorder = "1.4.3"
 
 zstd = { version = "0.9.0", optional = true }
 
+[target.'cfg(unix)'.dependencies]
+ctrlc = { version = "3", features = ["termination"] }
+
 [dev-dependencies]
 expect-test = "1.1.0"
 rstest = "0.11.0"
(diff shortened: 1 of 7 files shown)


dalcde:
There are other things we might want to do when we get terminated, e.g. wait for existing writes to complete. This might be very problematic for nassau where we open a file for a long time. We can also wait for all existing computations to complete, but that is harder to do in a uniform way, and may take quite a while before things actually terminate.

JoeyBF:
I like it! I might want to switch ctrlc for something like signal-hook in the future, to handle the other signals. FWIW, slurm kills processes with a SIGTERM instead of a SIGINT. Also, the grace period before issuing a SIGKILL is 30 seconds by default.

dalcde:
The termination feature makes it handle SIGTERM as well as SIGINT.

PR #95, opened 2022-05-19
Add coordinate abstractions

JoeyBF:
I saw Jason had interesting code that handles bidegrees and bigraded vector spaces, so I ported it here with some modifications. Upstreamed from jassch/AdamsSSMasseyProductCalculator.

It remains to modify the rest of the code to use this machinery, but I wanted to get comments first.


ext/crates/sseq/src/coordinates/bidegree.rs
```diff
@@ -0,0 +1,103 @@
+use std::fmt::{self, Display, Formatter};
+use std::ops::{Add, AddAssign};
+
+use super::iter::{ClassicalIterator, StemIterator};
+
+/// Type synonym for (s, t) bidegrees.
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Bidegree {
+    /// Homological degree
+    s: u32,
```
hoodmane, line 10: I think we ought to work towards representing homological degree as a i32 instead of a u32. Then we can handle things like duals and homological shifts.

ext/crates/sseq/src/coordinates/lattice.rs
```diff
@@ -0,0 +1,244 @@
+use std::cmp::Ordering;
+
+/// Partially ordered type with greatest lower bounds, or meets.
+pub trait MeetSemilattice: PartialOrd<Self> {
```
hoodmane, line 4: This file seems like it might add an unnecessary abstraction. Do you have a link to a place where this is used? Maybe split it off into a separate PR and merge the rest of everything first?

hoodmane (commented):
Generally looks reasonable to me. Certainly would help with bookkeeping. One thing that might be nice is to add a `PhantomData` type parameter to many of the structs so there would be a way to type check that a `BidegreeGenerator` came from the right structure.

ext/crates/sseq/src/coordinates/bidegree.rs
JoeyBF: I agree. That seems like a pretty intrusive change, so should I bundle it with this PR or open a new one?

ext/crates/sseq/src/coordinates/lattice.rs
JoeyBF: It's really only used outside of `ext-rs` itself. Jason's code uses it to figure out the region it needs to iterate over when computing Massey products

ext/crates/sseq/src/coordinates/bidegree.rs
hoodmane: > open a new one?

Yeah in a separate PR would be better.

dalcde:
Haven't read through the code closely, but it is probably easier to make the s/t fields of Bidegree pub

JoeyBF:
> Haven't read through the code closely, but it is probably easier to make the s/t fields of Bidegree pub

I didn't want to do that because we might want to change the internal representation at some point

JoeyBF:
I'll merge this PR next Friday if there aren't any more comments

hoodmane:
Thanks @JoeyBF!

JoeyBF:
No problem!

PR #96, opened 2022-05-20
Fix beta lints in ext

JoeyBF:
A new clippy lint was complaining about unnecessary allocations when appending to Strings using `format!`. There's still one failure in sseq/web_ext/sseq_gui/tests/test_3.py, where somehow `test_calpha` fails on line 42 but only in beta; stable and nightly are fine.

Edit: Apparently that test_calpha is pretty hit-or-miss because it passed just now...

### DIFF
--- a/ext/crates/algebra/src/algebra/adem_algebra.rs
+++ b/ext/crates/algebra/src/algebra/adem_algebra.rs
@@ -2,6 +2,9 @@
 
 use std::fmt;
 
+#[allow(unused_imports)]
+use std::fmt::Write as _; // Needed for write! macro for String
+
 use itertools::Itertools;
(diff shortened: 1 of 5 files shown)


dalcde:
You can avoid the allow(unused_import) by moving the import inside the mod test

JoeyBF:
Yep, totally didn't notice many of these were only used for tests. I was actually puzzled why it was an unused import in the first place since it does end up making the `write!` macro work, but that's just the `use super::*` at the top.

The `test_calpha` thing is still super flakey and I have no idea how to fix that. Maybe that should be a separate PR

dalcde:
I've fixed the flaky selenium tests

Issue #99, opened 2022-06-14
Use databases for saving


JoeyBF:
It would be more convenient to store resolutions using some kind of database, e.g. SQLite.

- Since we have (tens of millions of) very small files, a lot of the disk space is just filesystem overhead. Aggregating them would move us closer to solving the problem of using too much space (#76).
- This would make it easy to handle save files in an ACID way. I keep running into problems with using the very same files concurrently.

I would like to be assigned to this issue if that's ok. We can discuss the database scheme here or on Zulip.

JoeyBF:
Update: I looked into it and SQLite might be a bad choice. On one hand, very small resolutions are in the ideal size range for SQLite, but on the other hand, our record resolutions take up multiple TBs and require concurrent access by several hundred threads (e.g. to compute products) which is the opposite of ideal. With that said, something like PostgreSQL would work great with the large resolutions but would be massive overkill for the small ones.

Maybe we should abstract over the backing databases so we can choose at compile/run time?

dalcde:
I think sql just completely lacks ability to concurrently access from multiple processes (as opposed to threads), so having an interface that supports sqlite could be quite difficult

JoeyBF:
You're right. In that case we might want to have a new "database" feature that connects to postgres. That way we can keep the low-overhead save format for small resolutions

JoeyBF:
I'll have to add some methods in fp for saving and loading from sql, so I think it would be better if we finalize #53 beforehand, and possibly the upcoming matrix rewrite as well

PR #101, opened 2022-06-29
Add IO logging for Nassau quasi-inverses

JoeyBF:
I have a feeling the main bottleneck on the grid is IO, so I added a log entry. It also gives a more granular overview of the progress in a single bidegree, since there's one line for every signature now. I feel this is important now that every bidegree takes several hours. It's also future-proof w.r.t. #99, since we are generic over a `T: Write`

### DIFF
--- a/ext/src/utils.rs
+++ b/ext/src/utils.rs
@@ -615,7 +615,6 @@ mod logging {
         pub fn end(self, _msg: std::fmt::Arguments) {}
     }
 
-    #[repr(transparent)]
     pub struct LogWriter<T>(T);
 
     impl<T: Write> Write for LogWriter<T> {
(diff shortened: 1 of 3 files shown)


ext/src/utils.rs
```diff
@@ -572,9 +614,31 @@ mod timer {
 
         pub fn end(self, _msg: std::fmt::Arguments) {}
     }
+
+    #[repr(transparent)]
```
dalcde, line 618: Why is this needed?

JoeyBF: Just to make sure the newtype is really zero-cost. It's probably unnecessary, but I figured why not make sure

JoeyBF:
I'll also add that since this PR gives a much more verbose log than previously, it might be a good idea to start using something like `env_logger` to control the log level

dalcde:
I don't think this is the correct way to use repr(transparent); it should only be used for e.g. ABI guarantees.

JoeyBF:
I think I picked that up from a video somewhere. I removed it

PR #109, opened 2022-10-18
Fix all lints

JoeyBF:
It's been a few months so there were a few new lints to fix. Most importantly:

 - `OooTracker` is now a `BTreeSet` instead of a `Box<BTreeSet>`. This shouldn't make a difference since a `BTreeSet` is already heap-allocated.
 - `PPart` is now u32, no longer either u16 or u32 depending on features. This solves the numerous superfluous cast lints that clippy was upset about, and benchmarking shows that there is no measurable hit in performance.


JoeyBF:
See rayon-rs/rayon#952 for the miri test failure

JoeyBF:
#110 fixes miri

PR #113, opened 2022-11-07
Add Mahowald invariant example

aergus:
Here is an example that computes algebraic Mahowald invariants, showcasing `ResolutionHomomorphism` in particular.

### DIFF
--- a/ext/examples/mahowald_invariant.rs
+++ b/ext/examples/mahowald_invariant.rs
@@ -45,7 +45,11 @@
 //! [bruner--greenlees]: https://projecteuclid.org/journals/experimental-mathematics/volume-4/issue-4/The-Bredon-L%C3%B6ffler-conjecture/em/1047674389.full
 
 use algebra::{module::homomorphism::ModuleHomomorphism, AlgebraType};
-use ext::{chain_complex::ChainComplex, resolution_homomorphism::ResolutionHomomorphism, utils};
+use ext::{
+    chain_complex::{ChainComplex, FreeChainComplex},
+    resolution_homomorphism::ResolutionHomomorphism,
+    utils,
+};
 use fp::{matrix::Matrix, prime::TWO, vector::FpVector};
 
 use anyhow::Result;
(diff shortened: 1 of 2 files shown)


hoodmane:
Sorry for just leaving this here. We should probably merge this...

ext/examples/mahowald_invariant.rs
```diff
@@ -0,0 +1,201 @@
...
+
+        for (s, _, t) in s_2_resolution
+            .iter_stem()
+            .filter(|&(s, _, t)| p_k_resolution.has_computed_bidegree(s, t - 1))
+        {
+            let t_bottom = t + k as i32 - 1;
+            let bottom_s_2_gens = s_2_resolution.module(s).number_of_gens_in_degree(t_bottom);
+            let minus_one_s_2_gens = s_2_resolution.module(s).number_of_gens_in_degree(t);
+            let t_p_k = t - 1;
+            let p_k_gens = p_k_resolution.module(s).number_of_gens_in_degree(t_p_k);
```
JoeyBF, line 130:
```suggestion
            let bottom_s_2_gens = s_2_resolution.number_of_gens_in_bidegree(s, t_bottom);
            let minus_one_s_2_gens = s_2_resolution.number_of_gens_in_bidegree(s, t);
            let t_p_k = t - 1;
            let p_k_gens = p_k_resolution.number_of_gens_in_bidegree(s, t_p_k);
```

JoeyBF (approved):
It would be nice to have some tests, but I'm ok without too. Thanks for contributing!

JoeyBF:
I'll add that the lint error is just due to a new version of clippy and the selenium error is the same as in #121

aergus:
Thanks for the feedback. I'm trying to refactor the code to make it more testable and will comment here again when it's done (or I give up).

aergus:
I've added a test with a few cases (and "fixed" the incompatibility with the `nassau` feature).

JoeyBF:
Looks great to me! @hoodmane anything to add?

PR #114, opened 2022-11-22
Fix rust-analyzer

JoeyBF:
Rust-analyzer didn't understand which `apply_quasi_inverse` was being called, which is annoying. I also fixed some lints that show up when concurrency is enabled.


JoeyBF:
I'll add that the CI fail is an instance of la10736/rstest#169

PR #115, opened 2022-11-24
Optionally save homomorphisms and homotopies

JoeyBF:
I think it's a good idea to give the option of not saving homomorphisms and homotopies even when the source has a save directory. It makes quick operations faster since they are less IO-intensive. It will also make it easier to compute Massey products in the future, since right now computing homotopies is limited by IO, and they are only used once anyway

### DIFF
--- a/ext/src/chain_complex/chain_homotopy.rs
+++ b/ext/src/chain_complex/chain_homotopy.rs
@@ -43,7 +43,16 @@ impl<
         left: Arc<ResolutionHomomorphism<S, T>>,
         right: Arc<ResolutionHomomorphism<T, U>>,
     ) -> Self {
-        let save_dir = if left.source.save_dir().is_some()
+        Self::new_with_save(left, right, true)
+    }
+
+    pub fn new_with_save(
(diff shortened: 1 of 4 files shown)


JoeyBF:
Actually, now that I'm thinking about it a bit more, maybe it would be a better idea to have the option of not writing to disk but still reading what's already there

dalcde:
Do you still want to merge this and work on that separately? Also I think it would be better to raise/warn if save is true but there is no save file (maybe make it an enum with Yes/No/Auto?).

JoeyBF:
> Do you still want to merge this and work on that separately? Also I think it would be better to raise/warn if save is true but there is no save file (maybe make it an enum with Yes/No/Auto?).

That's a good idea, I implemented that with the `SaveOption` enum. I think it would be worth it to also give the option for primary and secondary resolutions, but that's more intrusive and that raises questions for ergonomics (e.g. what happens when we build a homomorphism with a nontrivial `save_option` between resolutions that also have nontrivial `save_option`s?)

PR #125, opened 2023-03-28
Fix rare UB in `once` crate when panicking

JoeyBF:
We're hitting some UB, in the form of segfaults and illegal instructions, and I tracked it down to that `unsafe` block. Some invariants aren't satisfied if we are panicking, so we just leak all the memory in that case

### DIFF
--- a/ext/Makefile
+++ b/ext/Makefile
@@ -75,5 +75,5 @@ examples/benchmarks/%-concurrent: dummy
 	fi
 
 miri:
-	cargo miri test -p once
+	MIRIFLAGS="-Zmiri-ignore-leaks" cargo miri test -p once
 	MIRIFLAGS="-Zmiri-ignore-leaks -Zmiri-disable-isolation" cargo miri test -p once --features concurrent
--- a/ext/crates/once/src/lib.rs
+++ b/ext/crates/once/src/lib.rs
@@ -169,16 +169,30 @@ impl<T> Drop for OnceVec<T> {
     fn drop(&mut self) {
         let len = self.len();
 
-        unsafe {
-            // The lock may be poisoned. Access is always safe because we have mutable reference,
-            // but if we can acquire the lock we want to drop the elements inside. If the lock is
-            // poisoned, then we are probably panicking so we don't care about memory leakage.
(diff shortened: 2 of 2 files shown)


dalcde:
I'm not convinced it is an issue with panicking per se; I think this
code path is rarely (never?) hit outside of panicking. I think it would
be better to look into the underlying issue more deeply.

JoeyBF:
By this code path you mean when the lock is poisoned? The thing is that I agree with the justifications for the unsafe block, I'm not sure what kind of UB is even possible except if we are somehow already in a partially deallocated state (which shouldn't be possible either!)

dalcde:
I mean dropping when ooo is non-empty

hoodmane:
@JoeyBF maybe we should just merge this for now? Since nobody seems to have the energy to look into this more carefully. Also, do we have any automation that runs miri? Presumably if we have tests for `OnceVec` that inject panics the runtime checks can tell us whether were are doing okay...

JoeyBF:
Strangely, adding a test that just pushes out of order then panics doesn't set off miri without this PR, and only reports a memory leak with it (as expected). Maybe merge this and open an issue to look into it more closely?

hoodmane:
Ideally it would be good to make a test that miri flags as undefined behavior without this change as part of this pr. But has the problem recurred? If you can't reproduce it and it doesn't happen anymore then maybe the change is not needed.

JoeyBF:
I believe it's still there. I'm not sure what I was working on at the time but when I find it again I'll make a minimal example and add it as a test. In the meantime I guess we can just leave this PR as-is

JoeyBF:
Closing: the bug was designed out rather than fixed.

This guarded `OnceVec::drop` with `thread::panicking()` to avoid a double-free in the manual `drop_in_place` loop over raw pages. **Master's `OnceVec` has no `Drop` impl at all** — it now lives in `once/src/once.rs`, rebuilt on `Grove`, and the raw-pointer teardown that was UB is gone entirely. Miri passes on master.

The `#[should_panic]` test here (`test_drop_panic`) was the durable part, since it existed to give Miri a panicking path to check. Worth re-adding against the current structures if that coverage is wanted — but it can't apply as written.

Issue #123, opened 2023-03-27
Use a finer directory structure for save directories


JoeyBF:
Right now we're having issues when the number of files in a directory is very large. On the Grid, operations on directories with more than ~50,000 files are very slow, and directories literally can't contain more than 1,000,000 files. We're hitting this limit for some computations.

I suggest we separate the various files by homological degree.

PR #127, opened 2023-04-04
Introduce finer save structure

JoeyBF:
This should solve #123

### DIFF
--- a/ext/pytests/test_secondary_job.py
+++ b/ext/pytests/test_secondary_job.py
@@ -2,7 +2,7 @@ import subprocess
 
 
 def folder_pairs(folder):
-    filename_parts = (f.stem.split("_") for f in folder.glob("*"))
+    filename_parts = (f.stem.split("_") for f in folder.glob("**/*") if f.is_file())
     return (
         (int(split[0].replace("m", "-")), int(split[1])) for split in filename_parts
     )
--- a/ext/src/chain_complex/chain_homotopy.rs
+++ b/ext/src/chain_complex/chain_homotopy.rs
@@ -49,7 +49,13 @@ impl<
             && !right.name().is_empty()
         {
             let mut save_dir = left.source.save_dir().clone();
-            save_dir.push(format!("massey/{},{}/", left.name(), right.name()));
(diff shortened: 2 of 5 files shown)


JoeyBF:
@hoodmane Any thoughts on this? This is getting relevant again as we're prepping for more computations soon.

I was thinking, since this isn't backwards-compatible with the current save format anyway (although fairly close), that this would also be a good time to rename the files themselves. For many purposes, it's more convenient to have `n_s_kind` than `s_t_kind`. One issue is that in theory stems can be negative, but there's simple fixes for that, e.g. printing `-i` as `mi`.

hoodmane:
Sorry for the super long delay in review. I'll try to take a look soon, but if you need it feel free to merge. It would be nice if we could get someone else to contribute/review changes. But it's hard.

hoodmane:
> printing -i as mi

Maybe `kind_-1_7` would be fine too.

JoeyBF:
> Sorry for the super long delay in review.

No worries! If it had been urgent I would have pinged you earlier but I get that other things get in the way sometimes.

> It would be nice if we could get someone else to contribute/review changes. But it's hard.

Agreed. It seems like the intersection of people who are proficient in both Rust and computational homotopy theory isn't very big, or maybe they prefer spending their time on their own projects.

> Maybe `kind_-1_7` would be fine too.

Yeah but I'm really uneasy about special characters like that in filenames. Also, I tried putting the kind first but it breaks one of the tests in a nontrivial way because it makes the filename harder to parse. If we look at segments delimited by '_', then the stem is either the second or third segment, because kinds sometimes have underscores in their name. It's also either the second to last or third to last, because sometimes we have an extra `_idx` tacked on at the end. I created #180 to test that out.

JoeyBF:
Closing in favour of #260 (ZarrV3 save rework).

This restructures `ext/src/save.rs` — directory layout, the #180 file rename, a TOCTOU fix on directory creation/deletion. #260 replaces that file's storage model wholesale, so this would conflict head-on and the migration it defines would be a format change to a format being retired.

PR #128, opened 2023-08-23
Fix most recent lints

JoeyBF:
Mostly syntactic changes, except I changed `module_construct_error.rs::test` to use `assert`s, and I changed the `PhantomData` in `sseq::Sseq` to make it Send and Sync

### DIFF
--- a/ext/Cargo.toml
+++ b/ext/Cargo.toml
@@ -4,6 +4,7 @@ version = "0.1.0"
 authors = [
     "Hood Chatham <hood@mit.edu>",
     "Dexter Chua <dexter@math.harvard.edu>",
+    "Joey Beauvais-Feisthauer <joeybf@wayne.edu>",
 ]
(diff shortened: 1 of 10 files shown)


ext/Cargo.toml
```diff
@@ -1,18 +1,23 @@
 [package]
 name = "ext"
 version = "0.1.0"
-authors = ["Hood Chatham <hood@mit.edu>", "Dexter Chua <dexter@math.harvard.edu>"]
+authors = [
+    "Hood Chatham <hood@mit.edu>",
+    "Dexter Chua <dexter@math.harvard.edu>",
+]
```
hoodmane, line 7: Maybe add yourself here?

JoeyBF: Ah yes, some toml changes are due to my autoformatter. Thanks for the suggestion, I'll add my name

PR #133, opened 2023-09-08
Introduce proptesting in `fp`

JoeyBF:
I wanted to try my hand at proptesting. Right now it only checks that row reduction turns a matrix built from a sequence of row operations applied to a matrix in RREF back to that same RREF. It's a bit more robust and it will be useful if (when) we mess around with the fp internals again

### DIFF
--- a/ext/crates/fp/tests/row_reduce.rs
+++ b/ext/crates/fp/tests/row_reduce.rs
@@ -2,6 +2,7 @@ use proptest::prelude::*;
 
 use fp::{matrix::Matrix, prime::ValidPrime};
 
+/// An arbitrary `ValidPrime`
 fn arb_prime() -> impl Strategy<Value = ValidPrime> {
     prop_oneof![
         Just(ValidPrime::new(2)),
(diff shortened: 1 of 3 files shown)


ext/crates/fp/tests/row_reduce.rs
```diff
@@ -0,0 +1,73 @@
...
+fn arb_prime() -> impl Strategy<Value = ValidPrime> {
+    prop_oneof![
+        Just(ValidPrime::new(2)),
+        Just(ValidPrime::new(3)),
+        Just(ValidPrime::new(5)),
+        Just(ValidPrime::new(7)),
+    ]
+}
+
+fn arb_tableau(rows: usize, cols: usize) -> impl Strategy<Value = Vec<usize>> {
```
hoodmane, line 14: Could you add a comment explaining the logic here? It's not completely obvious. Should we just assert `rows <= cols`? Arguably it doesn't make much sense otherwise.

ext/crates/fp/tests/row_reduce.rs
```diff
@@ -0,0 +1,73 @@
...
+        })
+        .chain(std::iter::repeat(vec![0; columns]).take(rows - tableau.len()))
+        .collect();
+
+        Matrix::from_vec(p, &row_vec)
+    }
+}
+
+prop_compose! {
+    fn arb_matrix_pair()
```
hoodmane, line 49: Maybe this could be given a more descriptive name like `reduced_nonreduced_pair`

hoodmane (approved):
Looks really nice. Thanks @JoeyBF! Property based testing catches some very weird bugs in my experience. It is unfortunate that it makes the CI less deterministic but I think it's very much worth it.

Please add comments to `arb_tableau`, `arb_row_pair`, and `arb_coeff_row_pair_seq` describing what they do so it's a bit easier to read. Also, could you add a brief explanation of the `prop_compose!` macro? It is pretty hard to tell from context what it does.

ext/crates/fp/tests/row_reduce.rs
JoeyBF: I agree, that's a better name

JoeyBF:
Thanks for the review! That made me realize some functions were unnecessary complicated, so I fixed those too

PR #134, opened 2023-09-11
Move `fp::vector` internals around

JoeyBF:
The diff is large but I didn't do anything other than move the vector-related code to its own module and split the impls into their own file. I feel like the code is easier to navigate that way and will be easier to modify in the future


hoodmane (approved):
Thanks @JoeyBF, indeed it looks like an improvement.

PR #136, opened 2023-09-25
Fix selenium but only partially

### LINKED ISSUE #121: Fix selenium test
There's a CI test that consistently fails right now, either by a timeout or a value error. This can't be caused by us, because commit ef6bc24 passed all tests, while the merged counterpart 16a0df9 failed

JoeyBF:
This fixes #121 but in an unsatisfying way. This seems to fix the flakiness of the selenium tests, but while inspecting the logs I noticed that the wasm/local/local-concurrent steps were actually all testing the wasm server. The servers compiled by cargo would exit when they noticed that port 8080 was busy, because the previous `make serve-wasm` was still running. Adding a `kill %1` to `ext.yaml`, so that the servers are stopped in between steps, only made more issues crop up. I feel like this should be fixed but as a separate PR

### DIFF
--- a/web_ext/sseq_gui/tests/test_s_2.py
+++ b/web_ext/sseq_gui/tests/test_s_2.py
@@ -120,6 +120,7 @@ def test_history(driver):
 
     timeout = 0.1
     while True:
+        print(f"Waiting for {timeout} seconds")
         time.sleep(timeout)
 
         try:
@@ -127,6 +128,7 @@ def test_history(driver):
                 file_contents = f.read()
                 if file_contents == "":
                     # The driver hasn't finished saving the file yet, wait a bit longer
+                    print("Waiting for driver to finish")
                     raise FileNotFoundError
                 driver.check_file("s_2.save", file_contents)
             break
--- a/web_ext/sseq_gui/tests/test_s_2.py
+++ b/web_ext/sseq_gui/tests/test_s_2.py
@@ -1,4 +1,6 @@
 import time
+from pathlib import Path
+
 from selenium.webdriver.common.by import By
 from selenium.webdriver.common.keys import Keys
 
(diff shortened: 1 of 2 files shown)


web_ext/sseq_gui/tests/test_s_2.py
```diff
@@ -120,11 +120,17 @@ def test_history(driver):
...
 
         try:
             with open(f"{driver.tempdir}/s_2.save") as f:
-                driver.check_file("s_2.save", f.read())
+                file_contents = f.read()
+                if file_contents == "":
+                    # The driver hasn't finished saving the file yet, wait a bit longer
+                    print("Waiting for driver to finish")
+                    raise FileNotFoundError
+                driver.check_file("s_2.save", file_contents)
```
hoodmane, line 133: What about:
```py
timeout = 0.1
save_path = Path(driver.tempdir) / "s_2.save"
while True:
   print(f"Waiting for {timeout} seconds")
   time.sleep(timeout)
   file_contents = save_path.read_text()
   if save_path.read_text():
      break
   timeout *= 2
driver.check_file("s_2.save", file_contents)
````

Comment (hoodmane) [reply web_ext/sseq_gui/tests/test_s_2.py:133]: what about:
```py
file_contents = save_path.read_text() if save_path.exists() else None
```

JoeyBF: I agree that's cleaner. Do we want to keep the max 10s for the timeout?

hoodmane: Yeah probably, but was there a max of 10s for the timeout before? I don't see any indication of that. Maybe `timeout = min(2*timeout, 10)`?

JoeyBF: It was on line 131, just after the end of the diff unfortunately. I put it in the suggested changes commit

hoodmane (approved):

PR #138, opened 2023-10-16
Fix regression when `SECONDARY_JOB` environment variable is set

JoeyBF:
It was introduced by me in #95 because the code preserved the `t` coordinate instead of the expected `n`


hoodmane:
Add a test?

JoeyBF:
> Could you add a test?

Since we want to test the `secondary` example itself, I figured I would write a bash script that runs it and checks for the existence and nonexistence of the proper files. I put that in the Makefile, and I checked that that test fails without the current PR

ext/bin/test_secondary_job.sh
hoodmane, line 1: I think this sort of bash script is a bit error prone. What about using pytest? E.g. a file like `test_secondary_job.py`:
```py
import subprocess


def folder_pairs(folder):
    return (f.stem.split("_") for f in folder.glob("*") if f.is_dir())


def test_secondary_job(tmp_path, monkeypatch):
    max_stems = 10
    goal_stem = 3
    monkeypatch.setenv("SECONDARY_JOB", str(goal_stem))
    subprocess.run(
        [
            "cargo",
            "run",
            "--example",
            "secondary",
            "--",
            "S_2",
            tmp_path,
            str(max_stems),
            "6",
        ],
        check=True,
    )
    for folder in [
        "augmentation_qis",
        "differentials",
        "kernels",
        "req_qus",
        "secondary_composites",
        "secondary_intermediates",
    ]:
        for s, t in folder_pairs(tmp_path / folder):
            assert t - s <= max_stems

    for s, _ in folder_pairs(tmp_path / "secondary_composites"):
        assert s == goal_stem

    for s, _ in folder_pairs(tmp_path / "secondary_intermediates"):
        assert s == goal_stem + 1
```

PR #141, opened 2023-11-23
Remove feature gate for serde

JoeyBF:
I'm not sure why they were there in the first place. The tests still pass if we just always implement Serialize/Deserialize


hoodmane (approved):
Seems reasonable to me. Thanks @JoeyBF!

PR #149, opened 2024-02-11
Implement affine subspaces

JoeyBF:
This will make it easier to handle Massey products in the future. Extends #148

### DIFF
--- a/ext/crates/fp/src/matrix/subspace.rs
+++ b/ext/crates/fp/src/matrix/subspace.rs
@@ -1,6 +1,6 @@
 use std::{
     io::{Read, Write},
-    ops::{Deref, DerefMut},
+    ops::Deref,
 };
 
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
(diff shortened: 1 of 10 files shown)


ext/crates/fp/src/matrix/subspace.rs
```diff
@@ -209,18 +214,34 @@ impl Subspace {
...
-        for row in self.matrix.iter().take(dim) {
-            if f.alternate() {
-                writeln!(f, "{row:#}")?;
-            } else {
-                writeln!(f, "{row}")?;
-            }
-        }
+        let mut rows = self.matrix.iter().take(dim);
+
+        let output = if f.alternate() {
```
hoodmane, line 234: What does `f.alternate()` mean?

hoodmane (approved):
Thanks!

ext/crates/fp/src/matrix/subspace.rs
JoeyBF: It formats the subspace in one line, to anticipate printing Massey products to `stdout`. I'll see if I can add a docstring.

hoodmane: Is there some choice of the name? What about calling it `one_line` or something a bit more explicit?

JoeyBF: It's a method on `std::fmt::Formatter` so we're stuck with that. In general for anything that implements Display you can print it normally using "{foo}" or use an alternate format with "{foo:#}".

JoeyBF: I added a doctest. Since it's part of the `Subspace` API it's also included in #148

PR #150, opened 2024-02-14
Introduce instrumentation using `tracing`

JoeyBF:
This is a more systematic solution than our previous homemade logging mechanism. I was worrying that the overhead might be too large, but the performance impact is only on a per-callsite basis. Even adding a few spans in the hottest code in `fp` doesn't measurably affect runtime.


hoodmane (approved):
Thanks!

PR #151, opened 2024-02-15
Ensure odd primes aren't defined when `odd-primes` is disabled

JoeyBF:
Since we moved to typed primes instead of `const u32`s in #145, it makes more sense to have 2 be the only prime that's defined when `odd-primes` is disabled. This gives us a type-level sanity check that all prime operations will be over 2, no matter which `P: Prime` is used as the generic.


hoodmane (approved):
Thanks!

PR #155, opened 2024-04-23
Various refactoring

adamnemecek:


hoodmane:
Looks reasonable to me. @JoeyBF any thoughts?

hoodmane:
Thanks @adamnemecek!

JoeyBF:
Looks good to me! I think I'll add a clippy directive to make sure we use `Self` consistently. Thanks!

PR #157, opened 2024-07-07
Set up `devenv` environment

JoeyBF:
I recently switched operating systems and I'm finding it much cleaner to use a dev environment than to install all the tooling system-wide. It also does wonders for reproducibility.

I chose `devenv` because of its convenience. The one remaining issue for me is the selenium test failing when running `devenv test` in `web_ext/sseq_gui`, but I just rely on CI for that.


hoodmane (approved):

hoodmane:
Thanks!

PR #161, opened 2024-07-25
Rename most "cocycle" -> "boundary"

JoeyBF:
I realized that `QueryCycleString` really queries a boundary, since it prints the value of the differential applied to the generator. The `differentials.rs` example also corroborates that


hoodmane (approved):
Thanks!

PR #162, opened 2024-08-01
Allow reading and writing in separate directories

JoeyBF:
This is handy if we are working on a big resolution and computing chain maps or homotopies. Computing maps and homotopies usually generates a ton of very small files, so it might be convenient to write them to some fast storage like an SSD or a ramdisk. Note that the way it is implemented means that it's impossible for a struct to read what it has previously written. This should not be a problem in practice.

Open problem: what should be the syntax to specify a split save directory in the examples? If we have a specific syntax we can just `impl FromStr for SaveDirectory` and make this feature available to examples without even changing our tests.

### DIFF
--- a/ext/src/chain_complex/chain_homotopy.rs
+++ b/ext/src/chain_complex/chain_homotopy.rs
@@ -55,7 +55,7 @@ impl<
                 .create_dir(save_dir.write().unwrap())
                 .unwrap();
 
-            left.source.save_dir().clone()
+            save_dir
         } else {
             SaveDirectory::None
         };
--- a/ext/src/chain_complex/chain_homotopy.rs
+++ b/ext/src/chain_complex/chain_homotopy.rs
@@ -1,7 +1,4 @@
-use std::{
-    path::{Path, PathBuf},
-    sync::{Arc, Mutex},
-};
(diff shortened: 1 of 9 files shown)


hoodmane:
Wouldn't this be better solved by using overlayfs?

JoeyBF:
It looks like you don't need root privileges, so it seems like it would work. In our case we would overlay a network drive with a local one, and we would have to do the same for every node that works on the resolution.

I'm hesitant though, because:
- I feel like it requires too much setup outside of `ext`, plus it only works on linux
- I was planning to make this a stepping stone towards more exotic save configurations, like using databases (#99), reading from multiple directories (if the resolution doesn't fit in a single directory), or writing the different types of files to different places

hoodmane:
Codewise looks good to me, so if you think it's a good approach feel free to merge. I'd think you could also accomplish those other things with a custom file system but if you think they are better implemented in rust then go ahead.

hoodmane (approved):

JoeyBF:
Thanks! I just noticed a small mistake I fixed for homomorphisms that I forgot to also fix with homotopies. I'll merge right after that!

PR #163, opened 2024-08-03
Update chart CI

JoeyBF:
I noticed the `chart` CI stopped working, and I found that `actions-rs/toolchain` is unmaintained. Thankfully, dtolnay has an (arguably better) action that does the same thing. This didn't end up fixing things, but after a bit more investigation, it turns out that removing `.rustup` from the cache does the trick


hoodmane (approved):
Thanks!

Issue #107, opened 2022-07-14
Allow checking integrity without loading resolution


JoeyBF:
I think my save might be corrupted but it's hard to know for sure, or which files exactly. It would be nice if there was a way to walk a save directory and verify the checksums individually

PR #164, opened 2024-08-11
Check integrity early for smaller files

JoeyBF:
This makes the save mechanism sturdier by catching malformed files early and overwriting them. Note that quasi-inverses are the only files that we stream in without reading completely, because they are so massive; other files fit in memory easily, so it's not a problem to load them early.

Compare with #107

### DIFF
--- a/ext/src/save.rs
+++ b/ext/src/save.rs
@@ -302,7 +302,7 @@ impl<T: Read> std::ops::Drop for ChecksumReader<T> {
 /// Open the file pointed to by `path` as a `Box<dyn Read>`. If the file does not exist, look for
 /// compressed versions. If `early_check` is true, we check the checksum before returning the file.
 fn open_file(path: PathBuf, early_check: bool) -> Option<Box<dyn Read>> {
-    fn do_early_check<T: Read>(path: PathBuf, mut reader: T) -> Option<Box<dyn Read>> {
+    fn do_early_check(path: PathBuf, mut reader: impl Read) -> Option<Box<dyn Read>> {
         let mut file_contents = Vec::new();
         let num_bytes = std::io::copy(&mut reader, &mut file_contents)
             .unwrap_or_else(|e| panic!("Error when reading from {path:?}: {e}"));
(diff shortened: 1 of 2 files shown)


JoeyBF:
Any feedback for this PR?

ext/src/save.rs
```diff
@@ -399,6 +433,17 @@ impl<A: Algebra> SaveFile<A> {
...
+    /// Whether we should load the file in memory and check the checksum before returning it. This
+    /// only returns false for quasi-inverses because they are our largest files by far. This is a
+    /// function of `SaveFile` and not just `SaveKind` because we may want to change the behavior
+    /// depending on the stem or some other heuristic.
+    fn should_check_early(&self) -> bool {
+        !matches!(
+            self.kind,
+            SaveKind::AugmentationQi | SaveKind::NassauQi | SaveKind::ResQi
+        )
+    }
```
hoodmane, line 445: This makes sense as an improvement over the status quo but there may be a better design of the file format that would allow checking only the section of the file that is used? For instance you could place a checksum every 4kb or imitate the zip file format. Another possibility is there might be an existing container format that is appropriate and that could make the checksums transparent to our code.

hoodmane (approved):

ext/src/save.rs
```diff
@@ -300,8 +300,35 @@ impl<T: Read> std::ops::Drop for ChecksumReader<T> {
...
+/// compressed versions. If `early_check` is true, we check the checksum before returning the file.
+fn open_file(path: PathBuf, early_check: bool) -> Option<Box<dyn Read>> {
+    fn do_early_check<T: Read>(path: PathBuf, mut reader: T) -> Option<Box<dyn Read>> {
+        let mut file_contents = Vec::new();
+        let num_bytes = std::io::copy(&mut reader, &mut file_contents)
+            .unwrap_or_else(|e| panic!("Error when reading from {path:?}: {e}"));
+        if num_bytes < 4 {
+            tracing::warn!("File {path:?} is too short to contain a checksum. Deleting file.");
+            std::fs::remove_file(&path)
+                .unwrap_or_else(|e| panic!("Error when deleting {path:?}: {e}"));
```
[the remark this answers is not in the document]
JoeyBF: Sure, that sounds like a good idea. Or maybe rename them with a ".old" suffix, and silently delete the previous .old version if there is one

ext/src/save.rs
JoeyBF: I looked around and didn't see a generic container that would do that for us, but one interesting option would be to always output zstd-compressed data. Zstd has options to do frame-level and block-level checksumming, and we already support reading compressed files.

Also, I just checked with our stem 200 data and, apart from the quasi-inverses, the biggest files only take a few MB, with the vast majority taking less than 1 kB. In fact, the first thing that the code does after opening those files is reading the entire contents, so we're really just prefetching

JoeyBF:
Closing in favour of #260 (ZarrV3 save rework), same reasoning as #127 — this modifies the `save.rs` that #260 replaces.

The underlying want is still live and better tracked as **issue #107** ("Allow checking integrity without loading resolution"). Whether zarr gives this for free (checksums per chunk) is worth checking while #260 is in flight.

PR #165, opened 2024-08-14
Add 130 tests for `fp::field`

JoeyBF:
I think this should cover every code path. The nice thing about modelling a mathematical structure, in this case a field, is you can check you're doing it right by checking the axioms. This is the ideal use case for proptest.

I added 13 tests that check the field axioms (and then some more) for `Fp<P>` and `SmallFq<P>` where `P` is any of: `P2`, `P3`, `P5`, `P7`, or `ValidPrime`. In this last case it checks the same primes as in the tests for `fp::vector`.

It also made me realize that there were a few methods that were missing from the API (`FieldElement::frobenius`, `Fp::element`, etc.) so I added them at the same time

### DIFF
--- a/ext/crates/fp/src/field/element.rs
+++ b/ext/crates/fp/src/field/element.rs
@@ -40,6 +40,14 @@ impl<F: FieldInternal> FieldElement<F> {
     pub(crate) fn val(self) -> F::ElementContainer {
         self.value
     }
+
+    pub fn inv(self) -> Option<Self> {
+        self.field.inv(self)
+    }
+
+    pub fn frobenius(self) -> Self {
+        self.field.frobenius(self)
+    }
 }
 
 // Allows us to access methods on `F::Element` directly
--- a/ext/crates/fp/src/field/fp.rs
+++ b/ext/crates/fp/src/field/fp.rs
@@ -190,7 +190,7 @@ mod tests {
 
         fn arb_elements<const N: usize>(
         ) -> impl Strategy<Value = (Fp<ValidPrime>, [FieldElement<Fp<ValidPrime>>; N])> {
-            crate::prime::tests::arb_prime().prop_flat_map(|p| super::arb_elements(p))
+            crate::prime::tests::arb_prime().prop_flat_map(super::arb_elements)
         }
 
         field_tests!();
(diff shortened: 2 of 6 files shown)


hoodmane (approved):
Thanks!

hoodmane:
It seems to me that it would be efficient to represent field elements as polynomials for small fields. Particularly in characteristic 2 I think this will be way faster than the logarithm approach. For example, for $F_4$ you can do:
```rs
let BIT_MASK: i64 = 0b01010101...;
let DEFINING_POLYNOMIAL: i64 = 0b11;
fn times_x(limb: i64) -> i64 {
   ((limb & (~BIT_MASK)) << 1) bitxor (((limb & BIT_MASK) >> 1) * DEFINING_POLYNOMIAL)
}

fn fma(limb: i64, elt: i64) {
   ((elt & 1) * limb) bit_xor ((elt >> 1) * times_x(limb))
}
```
And for $F_8$:
```rs
let BIT_MASK: i64 = 0b001001001001...;
let DEFINING_POLYNOMIAL: i64 = 0b011;
fn times_x(limb: i64) -> i64 {
   ((limb & (~BIT_MASK)) << 1) bitxor (((limb & BIT_MASK) >> 1) * DEFINING_POLYNOMIAL)
}

fn fma(limb: i64, elt: i64) {
   let xlimb = times_x(limb);
   let x2limb = times_x(xlimb);
   ((elt & 1) * limb) bit_xor (((elt >> 1) & 1) * xlimb) bit_xor ((elt >> 2) * x2limb)
}
```
For `F_9`:
```rs
let BIT_MASK: i64 = 0b000111000111...;
fn times_x(limb: i64) -> i64 {
   // In this case x^2 + 1 is a defining polynomial so *x just swaps the two fields and we don't need to reduce
   ((limb & (~BIT_MASK)) << 3) + (((limb & BIT_MASK) >> 3))
}

fn fma(limb: i64, elt: i64) {
   f3_reduce(((elt & 0b111) * limb) + f3_reduce((elt >> 3) * times_x(limb)))
}
```

JoeyBF:
You mean like vector operations would be faster? For large fields we definitely need polynomials because caching the Zech table is too expensive. But for field operations, on individual elements at least, it would be hard to beat lookup + add exponents for adding and multiply exponents for multiplying. Would it make sense to have this special logic in `SmallFq::{fma_limb, reduce}` instead?

PR #167, opened 2024-08-15
New secondary test

JoeyBF:
We shouldn't crash when extending a secondary homomorphism partially, but currently we are. This is stopping us from doing a large scale computation of secondary products.

Tagging our secondary expert @dalcde

### DIFF
--- a/ext/tests/secondary.rs
+++ b/ext/tests/secondary.rs
@@ -1,0 +1,41 @@
+use std::sync::Arc;
+
+use ext::{
+    resolution_homomorphism::ResolutionHomomorphism, secondary::*, utils::construct_standard,
+};
+use sseq::coordinates::Bidegree;
+
+#[test]
+fn boundary_has_boundary() {
+    ext::utils::init_logging();
+
+    // Construct resolution of S_2 to stem 15
+    let resolution = Arc::new(construct_standard::<false, _, _>("S_2", None).unwrap());
+    resolution.compute_through_stem(Bidegree::n_s(15, 9));
+
+    // Lift to secondary resolution
+    let res_lift = Arc::new(SecondaryResolution::new(Arc::clone(&resolution)));
+    res_lift.extend_all();
+
+    // Construct lift of "h0". This could ready be any element, but we choose "h0" for simplicity.
+    let hom = ResolutionHomomorphism::from_class(
+        "h0".to_string(),
+        Arc::clone(&resolution),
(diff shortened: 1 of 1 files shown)


JoeyBF:
Still reproduces on current master (`e30727e491`). The test compiles unchanged against today's APIs; the only adjustment needed is `let _ = ext::utils::init_logging();`, since that now returns a `Result`.

```
thread 'boundary_has_boundary' panicked at
crates/algebra/src/module/homomorphism/free_module_homomorphism.rs:162:22:
Index out of bounds: the length is 16 but the index is 16
```

Path to the panic:

```
SecondaryResolutionHomomorphism::compute_intermediates
  → get_intermediate → compute_intermediate
    → ModuleHomomorphism::apply → apply_to_basis_element
      → MuFreeModuleHomomorphism::output          <-- panics
```

### Diagnosis

`MuFreeModuleHomomorphism::output` guards its indexing with two asserts, and neither covers the failure:

```rust
pub fn output(&self, generator_degree: i32, generator_index: usize) -> &FpVector {
    assert!(generator_degree >= self.min_degree(), ...);
    assert!(generator_index < self.source.number_of_gens_in_degree(generator_degree), ...);
    &self.outputs[generator_degree][generator_index]   // line 162
}
```

The first checks the degree is not *below* `min_degree`. The second checks the index against the **source module's** generator count. Nothing checks `generator_degree < self.next_degree()` — that `outputs` was actually extended that far.

That is exactly the gap this test walks into. The resolution is computed through stem 15, but the homomorphism is only extended through stem 14, so `outputs.len() == 16` while the source module happily reports a non-zero generator count in degree 16. The second assert passes, and the `OnceBiVec` index then panics one past the end.

So partial extension is not itself unsupported — it is unchecked. The precondition "`outputs` is extended at least as far as the degree being asked for" is real but never asserted, which is why the failure surfaces as an out-of-bounds index deep inside `apply` rather than as a clear error at the call site in `compute_intermediate`.

Two separable questions:

1. **Should `output` assert `generator_degree < self.next_degree()`?** Cheap, and turns this into a message that names the actual problem. Worth doing regardless of 2.
2. **Should `compute_intermediates` avoid reaching past where the homomorphism was extended?** That is the real fix for the use case in the PR description (large-scale secondary products), and needs someone who knows the intended contract — @dalcde.

Found while triaging stale branches; the branch is being kept because of this PR.

PR #168, opened 2024-08-20
New `proptest` feature for `fp`

JoeyBF:
The `proptest::arbitrary::Arbitrary` trait lets us create values using a generic `any` free function. It would be really convenient to have an implementation of `Arbitrary` for types implementing `Prime` and `Field`, so that tests can just create values of that type out of thin air and focus on writing the contents of the tests. Let's focus on `Prime` for concreteness. Something like

```rust
#[cfg(test)]
mod tests {
    impl<P: Prime> proptest::arbitrary::Arbitrary for P {
        // stuff...
    }
}
```

isn't allowed because `P` is not constrained enough. One solution would be to use newtypes, but I found that messy and confusing. Another solution would be to add a `Prime: Arbitrary` constraint to the definition of `Prime`, but that exposes the proptest crate in our public API, and it would become a runtime dependency instead of a dev dependency.

As a compromise, I added a new feature that conditionally adds proptest support, using our `MaybeArbitrary` trait. This has the added benefit of exposing a mechanism for producing primes to other crates that depend on `fp` and also want to use proptest. This significantly simplifies the tests for our field types, and streamlines tests for vectors and row reduction.

Using these `Arbitrary` implementations, I was able to write a battery of tests that check the correct behavior of `FqVector` over every field and every prime. I'll submit this as a follow-up PR to keep the diff manageable.

### DIFF
--- a/ext/crates/fp/src/field/mod.rs
+++ b/ext/crates/fp/src/field/mod.rs
@@ -145,14 +145,7 @@ mod tests {
 
     pub(super) fn arb_elements<F: Field, const N: usize>(
     ) -> impl Strategy<Value = (F, [FieldElement<F>; N])> {
-        any::<F>().prop_flat_map(move |f| {
-            let elements: [_; N] = (0..N)
-                .map(|_| f.arb_element())
-                .collect::<Vec<_>>()
-                .try_into()
-                .unwrap();
-            (Just(f), elements)
-        })
+        any::<F>().prop_flat_map(move |f| (Just(f), std::array::from_fn(|_| f.arb_element())))
     }
 
     /// Test the field axioms and good behavior of the Frobenius endomorphism.
--- a/ext/crates/fp/src/field/mod.rs
+++ b/ext/crates/fp/src/field/mod.rs
@@ -24,11 +24,16 @@ pub trait Field: FieldInternal + Sized {
 
     fn zero(self) -> FieldElement<Self>;
     fn one(self) -> FieldElement<Self>;
+
+    #[cfg(feature = "proptest")]
+    fn arb_element(self) -> impl proptest::strategy::Strategy<Value = FieldElement<Self>>;
 }
 
 #[cfg(test)]
 mod tests {
-    use super::{Field, SmallFq};
+    use proptest::prelude::*;
+
+    use super::{element::FieldElement, Field, SmallFq};
     use crate::prime::P2;
 
     #[test]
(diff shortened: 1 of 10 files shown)


ext/crates/fp/src/field/mod.rs
```diff
@@ -138,16 +143,42 @@ mod tests {
...
+    pub(super) fn arb_elements<F: Field, const N: usize>(
+    ) -> impl Strategy<Value = (F, [FieldElement<F>; N])> {
+        any::<F>().prop_flat_map(move |f| {
+            let elements: [_; N] = (0..N)
+                .map(|_| f.arb_element())
+                .collect::<Vec<_>>()
+                .try_into()
+                .unwrap();
+            (Just(f), elements)
+        })
```
JoeyBF, line 155: I just found out about `std::array:from_fn`. This can just be
```rust
any::<F>().prop_flat_map(move |f| (Just(f), std::array::from_fn(|_| f.arb_element())))
```
but I'll make that change after the merge so it doesn't disrupt this PR and #169 too much

ext/crates/fp/src/field/smallfq.rs
```diff
@@ -286,6 +298,32 @@ impl<P: Prime> FieldInternal for SmallFq<P> {
...
 
+#[cfg(feature = "proptest")]
+impl<P: Prime> proptest::arbitrary::Arbitrary for SmallFq<P> {
+    type Parameters = ();
+    type Strategy = proptest::strategy::BoxedStrategy<Self>;
+
+    fn arbitrary_with(_args: Self::Parameters) -> Self::Strategy {
+        use proptest::prelude::*;
+
+        // Can't be a closure because the return value refenrences this function, and a closure
```
hoodmane, line 309:
```suggestion
        // Can't be a closure because the return value references this function, and a closure
```

hoodmane, line 309:
```suggestion
        // Can't be a closure because the return value references this function, and a closure
```

ext/crates/fp/src/prime/mod.rs
```diff
@@ -158,6 +162,20 @@ macro_rules! def_prime_static {
...
+
+        #[cfg(feature = "proptest")]
+        impl proptest::arbitrary::Arbitrary for $pn {
+            type Parameters = Option<NonZeroU32>;
+            type Strategy = proptest::strategy::Just<$pn>;
+
+            fn arbitrary_with(_max: Self::Parameters) -> Self::Strategy {
+                // This doesn't honor the max parameter, but that should be fine as long as the
+                // static primes are small enough and/or the max is large enough.
+                proptest::strategy::Just($pn)
```
hoodmane, line 174: You could presumably return an empty strategy if the max is too small, but it's probably not worth it.

ext/crates/fp/src/prime/primes_generic.rs
```diff
@@ -117,3 +117,50 @@ impl<'de> Deserialize<'de> for ValidPrime {
...
+    type Strategy = proptest::sample::Select<Self>;
+
+    /// An arbitrary `ValidPrime` in the range `2..(1 << 24)`, plus the largest prime that we
+    /// support. If `max` is specified, the primes are restricted to be less than `max`.
+    fn arbitrary_with(max: Self::Parameters) -> Self::Strategy {
+        use std::sync::OnceLock;
+
+        static TEST_PRIMES: OnceLock<Vec<ValidPrime>> = OnceLock::new();
+        let test_primes = TEST_PRIMES.get_or_init(|| {
+            // Sieve of erathosthenes
```
hoodmane, line 133:
```suggestion
            // Sieve of Eratosthenes
```

hoodmane (approved):
Looks good to me. I have a couple of spelling fixes -- maybe we should add a codespell lint...

ext/crates/fp/src/prime/mod.rs
JoeyBF: If I understand correctly it's actually impossible for a strategy not to give at least one value. Unless it counts everything as a reject, but then that causes local failures

JoeyBF:
> maybe we should add a codespell lint...

Haha that's fair, I'll see if I can add that!

PR #169, opened 2024-08-21
So many more tests in `fp`

JoeyBF:
A new suite of tests that check the correct behavior of the most important methods on `FqVector`s for every field type and prime type. This brings the total number of unit tests in `fp` to 442, but it only takes around 2-3 minutes to run through them all.

The tests made me realize that, in `FpVector`, the methods that take in two vectors check the compatibility of the primes at dispatch time. Since we don't have that check any more in `FqVector`, I needed to add those in. I also added a check to make sure that, whenever we pass in a slice of elements, they are also defined over the correct field. This should not affect the performance over small prime fields because equality between ZSTs compiles to a no-op.

Note that I also implemented the `from/to_bytes` etc. methods on `FqVector` instead of its wrapper so that the serialization test can check it directly.

The diff is large but #168 is a strict subset of the commits, so it's just an extension of that PR.

### DIFF
--- a/ext/crates/fp/src/field/element.rs
+++ b/ext/crates/fp/src/field/element.rs
@@ -40,6 +40,10 @@ impl<F: FieldInternal> FieldElement<F> {
         Self { field, value }
     }
 
+    pub fn field(&self) -> F {
+        self.field
+    }
+
     pub(crate) fn val(self) -> F::ElementContainer {
         self.value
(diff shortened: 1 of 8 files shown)


hoodmane (approved):
Thanks!

hoodmane (approved):
Thanks!

PR #174, opened 2024-10-07
Add `Arbitrary` implementation for other types in `fp`

JoeyBF:
I noticed a bug in `Subquotient::complement_pivots`, so I wanted to add a test to cover that. That led me to implement proptest's `Arbitrary` for several types. This PR also fixes the bug

### DIFF
--- a/ext/crates/fp/src/matrix/subquotient.rs
+++ b/ext/crates/fp/src/matrix/subquotient.rs
@@ -96,10 +96,8 @@ impl Subquotient {
 
     /// The pivot columns of the complement to the subspace
     pub fn complement_pivots(&self) -> impl Iterator<Item = usize> + '_ {
-        (0..self.ambient_dimension()).filter(|&i| {
-            !self.quotient.pivots().contains(&(i as isize))
-                && !self.gens.pivots().contains(&(i as isize))
-        })
+        (0..self.ambient_dimension())
+            .filter(|&i| self.quotient.pivots()[i] < 0 && self.gens.pivots()[i] < 0)
     }
 
     pub fn quotient(&mut self, elt: FpSlice) {
--- a/ext/crates/fp/src/matrix/subquotient.rs
+++ b/ext/crates/fp/src/matrix/subquotient.rs
@@ -239,6 +239,7 @@ pub mod arbitrary {
 #[cfg(test)]
 mod tests {
     use expect_test::expect;
+    use proptest::prelude::*;
 
     use super::*;
 
@@ -271,5 +272,15 @@ mod tests {
         .assert_debug_eq(&sq.reduce(FpVector::from_slice(p, &[2, 0, 0, 0, 0]).as_slice_mut()));
 
         assert_eq!(sq.gens().count(), 1);
+    }
+
+    proptest! {
+        #[test]
+        fn test_sum_quotient_gens_complement_is_ambient(sq: Subquotient) {
+            let quotient_dim = sq.zeros().dimension();
+            let gens_dim = sq.gens().count();
+            let complement_dim = sq.complement_pivots().count();
+            assert_eq!(quotient_dim + gens_dim + complement_dim, sq.ambient_dimension());
+        }
     }
 }
--- a/ext/crates/fp/src/matrix/subquotient.rs
+++ b/ext/crates/fp/src/matrix/subquotient.rs
@@ -184,6 +184,58 @@ impl Subquotient {
     }
 }
 
+#[cfg(feature = "proptest")]
+pub mod arbitrary {
+    use proptest::prelude::*;
+
+    use super::*;
+    use crate::matrix::subspace::arbitrary::SubspaceArbParams;
+    pub use crate::matrix::subspace::arbitrary::MAX_DIM;
+
+    #[derive(Debug, Clone)]
+    pub struct SubquotientArbParams {
+        pub p: Option<ValidPrime>,
+        pub dim: BoxedStrategy<usize>,
+    }
+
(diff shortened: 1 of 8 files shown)


ext/crates/fp/src/matrix/subspace.rs
```diff
@@ -298,3 +298,50 @@ impl std::fmt::Display for Subspace {
...
+                None => any::<ValidPrime>().boxed(),
+            };
+
+            (p, args.dim)
+                .prop_flat_map(move |(p, dim)| {
+                    Matrix::arbitrary_with(MatrixArbParams {
+                        p: Some(p),
+                        rows: (0..=dim + 1).boxed(),
+                        columns: Just(dim).boxed(),
+                    })
```
hoodmane, line 341: Is this likely to generate interesting subspaces? Almost all will be of full or nearly full rank. I'd think you should first pick the rank k, generate an arbitrary rank k * dim matrix, and then for each additional row pick a random linear combination of the first k rows. The last row can be left as all 0. Alternatively, it might be better to first pick a random row echelon form and then randomly pick the entries in the non-pivot columns.

ext/crates/fp/src/matrix/subquotient.rs
```diff
@@ -184,6 +184,58 @@ impl Subquotient {
...
+            (p, args.dim)
+                .prop_flat_map(|(p, dim)| {
+                    let sub = Subspace::arbitrary_with(SubspaceArbParams {
+                        p: Some(p),
+                        dim: Just(dim).boxed(),
+                    });
+                    let quotient = Subspace::arbitrary_with(SubspaceArbParams {
+                        p: Some(p),
+                        dim: Just(dim).boxed(),
+                    });
```
hoodmane, line 227: Doesn't the quotient have to be a sub of the sub here?

ext/crates/fp/src/matrix/subquotient.rs
```diff
@@ -96,10 +96,8 @@ impl Subquotient {
 
     /// The pivot columns of the complement to the subspace
     pub fn complement_pivots(&self) -> impl Iterator<Item = usize> + '_ {
-        (0..self.ambient_dimension()).filter(|&i| {
-            !self.quotient.pivots().contains(&(i as isize))
-                && !self.gens.pivots().contains(&(i as isize))
-        })
+        (0..self.ambient_dimension())
+            .filter(|&i| self.quotient.pivots()[i] < 0 && self.gens.pivots()[i] < 0)
```
hoodmane, line 100: Cool that this hit a bug =)

ext/crates/fp/src/matrix/subquotient.rs
hoodmane: It would be helpful to update this file with a comment explaining the invariants that sub and quotient passed in are supposed to satisfy and what additional invariants are satisfied when in normal form.

ext/crates/fp/src/matrix/subspace.rs
JoeyBF: The matrices themselves will be full rank the majority of the time, but they'll define a subspace of an ambient space of fixed dimension `dim`. The dimension of the subspace itself is the rank of the matrix, but because the number of rows ranges over `0..=dim+1`, I think we hit every subspace that way.

Is your point that we should test creating subspaces with highly rank deficient matrices? One of the invariants of `Subspace` is that it always contains a row-reduced matrix, so it'll be the same as creating a subspace with a smaller matrix and extending with rows of 0s. We could add a test that creates those deficient matrices and checks that the matrix does become row-reduced

ext/crates/fp/src/matrix/subquotient.rs
JoeyBF: I agree it's weird that quotient doesn't have to define a subspace of sub, but that's because a `Subquotient` containing `Subspace`s `gens` and `quotient` really represents the space (`gens` + `quotient`) / `quotient`, mathematically speaking.

This is documented at `Subquotient::from_parts`, but maybe it should be somewhere else too?

ext/crates/fp/src/matrix/subquotient.rs
JoeyBF: Yes! I'm quite happy with that haha

ext/crates/fp/src/matrix/subquotient.rs
hoodmane: How about at the top of the file?

ext/crates/fp/src/matrix/subspace.rs
hoodmane: That's good at least, but won't we still almost always end up in the dense Schubert cell for `Gr_{k, dim}` where the pivots lie along the main diagonal? I'd think what we'd want is to pick a random subset of `k` of the `dim` columns to be the pivot columns, and then fill the non pivot columns with random data.

JoeyBF: I simulated that by building an arbitrary matrix and forcing some pivots by filling initial segments with 0s. That should give more interesting subspaces

ext/crates/fp/src/matrix/subquotient.rs
JoeyBF: Good idea, I'll move it there and add a link to it at `Subquotient::from_parts`

JoeyBF:
Thanks! I added a few changes to export the new modules and update the integration test. I never realized we were only testing row-reduced matrices with 0s in the non-pivot columns, but now we're testing them all!

PR #175, opened 2024-10-21
Add `Matrix::transpose`

JoeyBF:
`Matrix` had no transpose. This adds one, blocked at `p = 2`.

Entry `(i, j)` of a `p = 2` matrix is bit `j % 64` of limb `j / 64` of row `i`, so a 64-square block
of entries is exactly one limb from each of 64 consecutive rows. `blas::block::transpose_lanes`
moves such a block with six masked shift-and-XOR passes — the recursive delta swap of Hacker's
Delight 7-3 — rather than the 4096 single-bit extractions an entry-at-a-time transpose performs.

Two things beyond the swap itself matter about as much again:

- **Filling cache lines.** Each source block contributes one limb to each of 64 destination rows, so
  a block at a time touched 64 cache lines to deposit 8 useful bytes each. Taking 8 blocks at a time
  makes those contributions 8 consecutive limbs of each destination row — one whole line.
- **One block per lane.** Every lane then runs the same swap on the same row indices, so nothing
  moves between lanes and the inner loop is a straight elementwise operation. This is adapted from
  [mkarppa/matmul](https://github.com/mkarppa/matmul) (MIT; Karppa & Kaski,
  [arXiv:1909.01554](https://arxiv.org/abs/1909.01554)), which transposes bit blocks over `uint4`
  rather than one word at a time; one block per lane is the CPU form of that width.

`Matrix::transpose` works an 8×8 tile of blocks through a 32 KB scratch, so both trips to main
memory are contiguous and the shuffling stays in L1. Odd primes keep the entry-at-a-time path.

Against that path, on square matrices:

| size | blocked | naive | speedup |
|------|---------|-------|---------|
| 512² | 0.008 ms | 1.23 ms | 156× |
| 1024² | 0.036 ms | 4.46 ms | 125× |
| 2048² | 0.111 ms | 18.90 ms | 170× |
| 4096² | 0.499 ms | 90.20 ms | 181× |
| 8192² | 2.162 ms | 388.02 ms | 180× |

Effective bandwidth is flat at 7.4–9.4 GB/s across that range. Measured on one dev box; it is a
cache-sensitive routine, so expect the numbers to move with the machine — beyond ~8192² the working
set leaves L3 and throughput falls off.

## Tests

- `transpose_lanes` against an entry-at-a-time oracle: random blocks, involution, and every
  single-bit block.
- `Matrix::transpose` under proptest at all primes (via the existing `Arbitrary for Matrix`), plus
  64 hand-picked shapes straddling block boundaries in each dimension independently — 1, 63, 64, 65,
  127, 128, 129, 200 in each axis.

## History

This branch previously held a naive entry-at-a-time transpose. It has been rewritten rather than
built on: the old implementation predates the `Matrix` rewrite and assumed `vectors: Vec<FpVector>`,
where `Matrix` now stores `data: AVec<Limb>` with a stride, which is what makes the blocked form
possible at all.

Split out of the work in SpectralSequences#285; the GPU half is a separate PR and does not depend on
this one.

## Summary by CodeRabbit

* **New Features**
  * Added matrix transposition support.
  * Binary-field matrices now use an optimized transposition approach for improved performance.
  * Transposition works correctly across supported matrix sizes and can be applied repeatedly.

* **Tests**
  * Added coverage for boundary dimensions, repeated transposition, single-bit behavior, and comparisons against a reference implementation.

### DIFF
--- a/ext/crates/fp/src/blas/block.rs
+++ b/ext/crates/fp/src/blas/block.rs
@@ -236,3 +236,108 @@ mod arbitrary {
         }
     }
 }
+
+/// Transpose the 64-square bit block held in each lane, all lanes at once.
+///
+/// `block[i][p]` is row `i` of lane `p`'s block; on return row `j` of lane `p` holds what was bit
+/// `j` of each of that lane's rows.
+///
+/// The recursive delta swap of Hacker's Delight 7-3: each round exchanges the off-diagonal
+/// quadrants of every sub-block at the current scale, so a full 64-square transpose costs six
+/// masked passes rather than the 4096 single-bit extractions an entry-at-a-time transpose performs.
+/// Every lane runs the same swap on the same row indices, so nothing moves between lanes and the
+/// inner loop is a straight elementwise operation over `LANES` limbs — the shape a vector unit
+/// wants, and the reason to transpose a panel of blocks together rather than one at a time.
+#[inline]
+pub fn transpose_lanes<const LANES: usize>(block: &mut [[Limb; LANES]; 64]) {
+    let mut s = 32;
+    // Selects the columns whose index has bit `s` clear: the left half of each 2s-wide group.
+    let mut m: Limb = !0 >> 32;
+    while s != 0 {
+        let mut k = 0;
+        while k < 64 {
+            // `k` never has bit `s` set, so `k` and `k | s` are the upper and lower row halves of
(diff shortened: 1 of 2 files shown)


hoodmane (approved):
It looks fine. But do we need it for something?

PR #176, opened 2024-10-30
2024ify

JoeyBF:
This ensures that the code is ready to be updated to the 2024 edition of Rust when it comes out in a few months. For us, the main pain point was the fact that `gen` is now a reserved keyword, so we can't use it to name generators. The way `cargo fix --edition` fixes this is by using the raw identifier `r#gen`, but I would rather not use raw identifiers. I renamed them all `g` instead, but I'm open to other names


hoodmane (approved):
Thanks!

PR #180, opened 2025-01-25
Rename saved files

JoeyBF:
We now use (stem, homological degree) notation instead of (homological degree, internal degree) in the names of files, which makes them easier to inspect. To avoid dashes in filenames, we replace minuses by the character `m`.

This is the change that I mentioned in https://github.com/SpectralSequences/sseq/pull/127#issuecomment-2613829194.

### DIFF
--- a/ext/pytests/test_secondary_job.py
+++ b/ext/pytests/test_secondary_job.py
@@ -2,9 +2,9 @@ import subprocess
 
 
 def folder_pairs(folder):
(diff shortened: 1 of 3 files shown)


hoodmane (approved):
Thanks!

JoeyBF:
> Thanks!

My pleasure! I changed the target branch to `save3.0` so we can push breaking changes all at once.

PR #182, opened 2025-02-02
Ensure save paths are absolute

JoeyBF:
This makes logging more reliable. Since enum fields are also constructors, we can't uphold that invariant in a `new` function. Instead, we ensure the paths are absolute at the type level with a thin wrapper.


hoodmane (approved):
Thanks!

PR #183, opened 2025-02-02
Better IO-related logs

JoeyBF:
This replaces `ChecksumWriter` by a writer that does more things. It does take care of checksums but it also logs the number of bytes written and the amount of time it was open.

This essentially does the job of `LogWriter`, so I removed that too.

### DIFF
--- a/ext/src/nassau.rs
+++ b/ext/src/nassau.rs
@@ -41,7 +41,6 @@ use sseq::coordinates::Bidegree;
 use crate::{
     chain_complex::{AugmentedChainComplex, ChainComplex, FiniteChainComplex, FreeChainComplex},
     save::{SaveDirectory, SaveKind},
-    utils::LogWriter,
 };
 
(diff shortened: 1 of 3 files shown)


JoeyBF:
Now that I'm thinking about it more, maybe we should just keep `LogWriter`, have *it* do the data/timing logging on drop, and just return a `LogWriter<ChecksumWriter>` when opening a file.

JoeyBF:
I'll mark this as a draft until we figure out whether we should have one save wrapper or next the two existing wrappers.

JoeyBF:
Closing as stale.

Two of the three commits (`Remove LogWriter`, `Upgrade save file writer to log more data`) touch `save.rs` and `nassau.rs` in ways #260 supersedes. The third is the #180 rename shared with #127.

If the logging improvements are wanted independently of the save format, they're small enough to redo against whatever #260 lands.

Issue #184, opened 2025-03-03
Add `Cohomology` abstraction layer


JoeyBF:
Many of the examples work at the level of the resolution itself, but really are all about the underlying Ext of whatever module in question. I suggest we add a struct that encapsulates a resolution but only exposes its cohomology (after taking Hom(-, $\mathbb{F}_2$)). Or maybe a struct that a resolution would create but that would not have to itself carry the resolution around.

In fact, it may be useful to have a homotopy group struct, since for any spectrum M, its cohomology is the homotopy of $C\lambda \otimes M$ as a synthetic spectrum. This ties in nicely with the secondary machinery that essentially works in $C\lambda^2$-modules instead of $C\lambda$.

PR #185, opened 2025-03-06
Add `OrderedBidegree` struct

JoeyBF:
I'm using this as part of some work on #184 but I feel it's natural to split it into a separate PR


hoodmane (approved):
Didn't read super carefully but it sounds quite reasonable.

JoeyBF:
Thanks! It just adds an API so we can always tweak it later

PR #186, opened 2025-03-06
Use `i32` for homological degree

JoeyBF:
This enables a few things.
- We can now reason about duals, which are negatively graded, even though this is completely untested and probably broken as-is.
- We can also use `Bidegree`s to index full-page spectral sequences, like in the `sseq` subcrate or the webserver.
- We can now store bidegrees as `[i32; 2]`, to allow for generalizing to `[i32; N]` for multi-graded objects in the future.


hoodmane (approved):
Nice also to see how many casts this removes.

JoeyBF:
I also found it very satisfying! Also because many fields went from `OnceVec` to `OnceBiVec`, we got rid of the usize casts too

PR #190, opened 2025-03-10
Integrate bidegrees into `sseq::charting`

JoeyBF:
Continuing the train of thought from #188 and #189, this brings it all together.


hoodmane (approved):
Thanks! looks quite a bit tidier.

Issue #187, opened 2025-03-08
Use `BTreeMap` backing for structs in `sseq`


JoeyBF:
I feel that replacing the current
```rust
pub struct DenseBigradedModule {
    dimensions: OnceBiVec<OnceBiVec<usize>>,
    min_y: i32,
}
```
with something like
```rust
pub struct DenseBigradedModule {
    dimensions: BTreeMap<OrderedBidegree<ByStem>, usize>,
}
```
would make sense. If we go ahead and refactor the `sseq` API to use `Bidegree`s throughout, it would make accessing data easier. I suppose we also reduce allocations and have one less indirection.

Similarly, I suggest we replace
```rust
pub struct Sseq<P: SseqProfile = Adams> {
    differentials: BiVec<BiVec<BiVec<Differential>>>,
    permanent_classes: BiVec<BiVec<Subspace>>,
    page_data: BiVec<BiVec<BiVec<Subquotient>>>,
    invalid: BiVec<BiVec<bool>>,
    // ...
}
```
with
```rust
pub struct Sseq<P: SseqProfile = Adams> {
    differentials: BTreeMap<OrderedBidegree<ByStem>, BiVec<Differential>>,
    permanent_classes: BTreeMap<OrderedBidegree<ByStem>, Subspace>,
    page_data: BTreeMap<OrderedBidegree<ByStem>, BiVec<Subquotient>>,
    invalid:  BTreeMap<OrderedBidegree<ByStem>, bool>,
    // ...
}
```

In fact, now that I write this all out, it might be even more convenient to use some helper struct
```rust
pub struct Bigraded<T>(BTreeMap<OrderedBidegree<ByStem>, T>)
```
everywhere.

hoodmane:
I think this would be bad for lookup speed. You should actually profile it before doing anything like this.

JoeyBF:
> I think this would be bad for lookup speed. You should actually profile it before doing anything like this.

I agree that I should be careful about that. What kind of benchmark do you think I should add? I suppose the underlying question is what kind of operation on a spectral sequence I should optimize for, which is unclear to me until I implement things further

hoodmane:
Well my suspicion is that if you did this for our Algebra data structure (for example) and then ran a normal minimal resolution, you'd see a significant slow down.

hoodmane:
I don't think this should be hard to detect or particularly sensitive to the workload.

JoeyBF:
[...]

PR #192, opened 2025-03-25
Overhaul of the `once` crate

JoeyBF:
This almost completely overhauls the `once`. We keep `OnceVec` and `OnceBiVec` and their API, but they are now backed by a more well-behaved data structure I call a `Grove`, for "grow vector". Bikeshedding welcome for the name.

The new `MultiIndexed<K, V>` struct is what I had in mind in https://github.com/SpectralSequences/sseq/issues/187#issuecomment-2715388602. As the criterion benchmark shows, its performance is comparable to nested `OnceBiVec`s for insertion, sometimes even faster. It also offers more flexibility like natively out-of-order insertions and arbitrarily negative coordinates.

I added doctests, unit tests, proptests and tests using [loom](https://github.com/tokio-rs/loom/) to test against strange thread scheduling. Also probably too much documentation. Sorry for the large diff


JoeyBF:
Yeah I totally agree with you. Hopefully enough of it is docs and tests to make it skimmable.

ext/crates/once/benches/criterion/benchable_impl.rs
```diff
@@ -0,0 +1,268 @@
...
+        let layer4 = OnceBiVec::new(min[4]);
+        layer3.push_checked(layer4, min[3]);
+        layer2.push_checked(layer3, min[2]);
+        layer1.push_checked(layer2, min[1]);
+        layer0.push_checked(layer1, min[0]);
+        layer0
+    }
+
+    fn push_checked(&self, coords: [i32; 5], value: T) {
+        if let Some(layer1) = self.get(coords[0]) {
```
hoodmane, line 202: Let's negate and unnest this as well so we get linearly much code per layer instead of quadratic:
```rs
    if self.get(coords[0]).is_none() {
        let layer1 = OnceBiVec::new(coords[1]);
        self.push_checked(layer1, coords[0]);
    }
    let layer1 = self.get(coords[0]).unwrap();
    if layer1.get(coords[1]).is_none() {
        let layer2 = OnceBiVec::new(coords[2]);
        self.push_checked(layer2, coords[1]);
    }
    ...
```
Or better yet, what about adding a `get_or_insert()` API and this function could be:
```rs
    self.get_or_insert(coords[0], || OnceBiVec::new(coords[1]))
        .get_or_insert(coords[1], || OnceBiVec::new(coords[2]))
        .get_or_insert(coords[2], || OnceBiVec::new(coords[3]))
        .get_or_insert(coords[3], || OnceBiVec::new(coords[4]))
        .get_or_insert(coords[4], || OnceBiVec::new(coords[5]))
```

ext/crates/once/src/grove/mod.rs
```diff
@@ -0,0 +1,970 @@
...
+
+use std::num::NonZero;
+
+use block::Block;
+
+use crate::std_or_loom::sync::atomic::{AtomicI32, AtomicUsize, Ordering};
+
+const MAX_NUM_BLOCKS: usize = 32;
+
+/// An insert-only sparse vector with pinned elements and geometrically growing capacity.
```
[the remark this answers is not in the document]
JoeyBF: I just left it like that because that's what `OnceVec` was doing under the hood already. It also makes computing the size of the next block easy because it's just a power of two. I'll experiment with polynomial growth in some benches and get back to you.

ext/crates/once/benches/criterion/benchable_impl.rs
```diff
@@ -0,0 +1,268 @@
...
+            let layer2 = OnceBiVec::new(coords[2]);
+            layer2.push_checked(layer3, coords[3]);
+            let layer1 = OnceBiVec::new(coords[1]);
+            layer1.push_checked(layer2, coords[1]);
+            self.push_checked(layer1, coords[0]);
+        }
+    }
+
+    fn get(&self, coords: [i32; 4]) -> Option<&T> {
+        if coords[0] >= self.min_degree() && coords[0] < self.len() {
```
hoodmane, line 160: Let's unnest this too.

JoeyBF:
Thank you! I really did have fun digging into lock-free data structures :smile:

PR #195, opened 2025-06-26
Add Quickstart section to README

ebelmont:
I wrote this section for the README after having conversations with several people who asked whether sseq had a certain feature or how to get started on a basic level. The goal is to make this more accessible to homotopy theorists who mostly want to make an Ext chart for their favorite module.

I talked to Joey about this, and also had someone successfully install sseq on windows using a version of these instructions, but who ran into the "gotchas" I listed below step 5 in the installation section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ allows us to work compute in the derived category of said algebra. The primary
 purpose is to compute the classical Adams E2 page by computing Ext over the
 Steenrod algebra.
 
+** To get started computing Ext over the Steenrod algebra, read the Quickstart [here](ext/README.md).**
+
 2. `web_ext`
 Web interfaces to `ext`. There are two subprojects at the moment:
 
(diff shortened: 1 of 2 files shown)


ebelmont:
It just occurred to me that this might be better on the top-level README, not the one for `ext`, because people will probably not see it otherwise. I'll leave the rest up to you.

hoodmane:
Just trying cloning myself and I notice clone time is a bit slow -- I think we had some binary assets checked in at some point but then removed them. We should probably suggest a depth 1 clone.

JoeyBF:
It would require a lot of files at least given the number of architectures Rust supports and the number of features we may want to enable or not (logging, zstd, concurrent, nassau, odd-primes, etc.). Plus we compile with `-C target-cpu=native` by default, so we would lose some performance, e.g. SIMD.

hoodmane:
Yeah I guess we still have WebAssembly for the quick start / no install needed version.

PR #197, opened 2025-07-04
Disable Selenium test for now

JoeyBF:
The switch from Ubuntu 20.04 which reached EOL to Ubuntu 24.04 broke our selenium tests. I don't have the bandwidth to deal with that right now, but at least the webserver jobs still work. Since `deploy` depends on `selenium`, that also means that we won't be deploying new versions until this gets fixed. I feel it's ok since the webserver hasn't changed in years.

### DIFF
--- a/.github/workflows/ext.yaml
+++ b/.github/workflows/ext.yaml
@@ -179,7 +179,9 @@ jobs:
           path: web_ext/sseq_gui/dist/
 
   selenium:
-    if: ${{ github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository }}
+    # TODO Someone please figure out how to make the selenium tests work on Ubuntu 24.04
+    if: false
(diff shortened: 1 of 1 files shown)


hoodmane:
It's failing to find the firefox driver, maybe it can find the chrome driver? Let's try just commenting out line 54:
https://github.com/SpectralSequences/sseq/blob/master/web_ext/sseq_gui/Makefile?plain=1#L54

JoeyBF:
No it gives the same error, just for the chrome driver: https://github.com/JoeyBF/sseq/actions/runs/16076913009/job/45374256882

What's extra strange is that this uses selenium 4.12, and since 4.6 it should [download the correct driver automatically](https://www.selenium.dev/documentation/webdriver/troubleshooting/errors/driver_location/)

JoeyBF:
Resolved in #208

PR #199, opened 2025-07-21
Add iteration and standard impls in `once` crate

JoeyBF:
It was getting annoying that `MultiIndexed` didn't implement Debug, Clone, PartialEq, Eq, Hash, since any structs that contain one would not be able to derive them. The main bottleneck was being able to iterate over the entire tree.

I implemented iterators and updated the criterion bench to make sure the performance is close to optimal. I also implemented all the standard traits for `MultiIndexed`, `KdTrie`, `Grove` and `TwoSidedGrove`. I added proptests to check for correctness too.

### DIFF
--- a/ext/crates/once/Cargo.toml
+++ b/ext/crates/once/Cargo.toml
@@ -15,6 +15,7 @@ maybe-rayon = { path = "../maybe-rayon" }
 
 [dev-dependencies]
 criterion = "0.5"
+expect-test = "1.1.0"
 pprof = { version = "0.14", features = ["criterion", "flamegraph"] }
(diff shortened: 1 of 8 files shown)


JoeyBF:
For future reference, I did try unwrapping the new methods in `benchable_impls.rs`, so that the impl for `Benchable<K, T>` would call the method for `Benchable<K-1, T>` in a `flat_map`. This makes me very confused as the unwrapped version is significantly faster for large K (~3x), but also significantly slower for K = 2 (~4x). I'll leave the implementation as-is because I feel it's the most representative for uses outside of this crate.

PR #208, opened 2025-10-27
Change Selenium error

JoeyBF:
After more debugging than I would like to admit, I think Selenium works again now. The issue is that now the flaky `calpha` test seems to fail consistently, but we should open a different PR to fix that.


JoeyBF:
Thanks, I wasn't aware of `xfail`!

PR #209, opened 2025-10-29
Determine SIMD at runtime

JoeyBF:
This way we support all architectures that Rust supports, but we dispatch to a SIMD specialization depending on which instruction sets are available _at runtime_. This is possible thanks to the `target_feature(enable = "...")` attribute. I will extend the SIMD capabilities in #207.


JoeyBF:
In an HPC setting for example, it makes it easier to generate one binary that we can use on all nodes without loss of performance.

In general, I'm planning to implement some of the BLAS API (as in #207), and the industry standard is to bundle a number of kernels and do runtime dispatch.

PR #245, opened 2026-06-25
Catch wasm worker panics and surface them as error dialogs

hoodmane:
With panic=unwind, wasm-bindgen rethrows Rust panics as JS exceptions out of the exported run() method. Previously these escaped uncaught in the worker and were invisible to the user. Wrap the run() calls in the sseq and resolution workers in try/catch and forward the failure as an Error action message, which the frontend already renders as a dialog.

Add a Node unit test (make test-wasm-js) that loads the worker scripts with mocked wasm globals, makes run() throw, and asserts an Error message is posted. Wire it into the webserver CI job.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling in background worker tasks so failures are reported back to the app instead of failing silently.
  * Added clearer error messages when worker execution encounters unexpected issues.

* **Tests**
  * Added automated coverage for worker error handling to verify both successful runs and failure reporting.
  * Included the new test in the existing workflow so it runs during continuous integration.

### DIFF
--- a/web_ext/sseq_gui/wasm/sseq_worker.js
+++ b/web_ext/sseq_gui/wasm/sseq_worker.js
@@ -6,10 +6,26 @@ const promise = wasm_bindgen("./sseq_gui_wasm_bg.wasm").catch(console.error).the
     self.sseq = Sseq.new((m) => self.postMessage(m));
 });
 
+function reportPanic(err) {
+    const message =
+        err && err.stack ? err.stack : `${err}`;
+    self.postMessage(
+        JSON.stringify({
+            recipients: [],
+            sseq: 'Main',
+            action: { Error: { message: `Panic in sseq worker:\n${message}` } },
+        }),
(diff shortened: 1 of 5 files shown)


web_ext/sseq_gui/wasm/sseq_worker.js
```diff
@@ -6,10 +6,26 @@ const promise = wasm_bindgen("./sseq_gui_wasm_bg.wasm").catch(console.error).the
...
+    const message =
+        err && err.stack ? err.stack : `${err}`;
+    self.postMessage(
+        JSON.stringify({
+            recipients: [],
+            sseq: 'Main',
+            action: { Error: { message: `Panic in sseq worker:\n${message}` } },
+        }),
+    );
+}
```
JoeyBF, line 19: Can we unify this with the other reportPanic?

PR #246, opened 2026-06-25
Add try_new variants for panicking constructors

hoodmane:
Add fallible variants for the remaining module constructors and FD builder methods that panic on bad preconditions.

- HomModule::try_new: Err when target is not bounded above
- QuotientModule::try_new: Err on too-low truncation / i32 overflow
- HomPullback::try_new: Err on inconsistent source/target/map wiring


JoeyBF (approved):
Looks good!

PR #251, opened 2026-06-29
Bit-sliced storage for finite-field vectors

JoeyBF:
## Summary

Replaces the **packed-limb** representation of finite-field vectors in the `fp`
crate with a **bit-sliced (bit-plane)** layout. Instead of packing each element
into `ceil(log2(p*(p-1)))` consecutive bits of one limb, a group of 64 elements
is stored across `k = ceil(log2 p)` limbs (planes): plane `j` holds bit `j` of
all 64 elements. Addition and scalar multiplication become branch-free boolean
circuits over the planes that operate on 64 elements at once, with **no separate
reduction step**.

F2 is the `k = 1` instance of this layout and is byte-identical to the old
packed representation, so the SIMD / m4ri / BLAS F2 fast paths are untouched.

## Acknowledgments

The idea of bit-slicing finite-field vectors, and the optimized 6-gate F3
addition circuit, were both contributed by **Carl McTague**
(carl.mctague@bc.edu, [@mctague434](https://github.com/mctague434)). He is
credited as a co-author on the relevant commit and in the source.

## Motivation

`add`/`scale` are the hot path of Gaussian elimination and resolution
computation. The old path did a fused multiply-add producing non-reduced limbs
followed by a separate SWAR `reduce`. The bit-sliced circuits remove the reduce
step entirely and were validated by benchmark before the refactor:

- Specialized circuits win for the tuned small primes (F3 ~2.5× add / ~1.75×
  scale vs packed at length 100k).
- A **const-generic `K`** dispatch (exactly-sized stack arrays, unrolled
  per-prime loops) made the generic kernel beat packed for *every* prime
  (add: ~3× at p=3 up to ~24× at p=251; scale: ~1.1× up to ~36×).

## What changed

- **Group-layout seam in `FieldInternal`** — `entries_per_group`,
  `limbs_per_group`, `group_of`/`lane_of`, `gather`/`scatter`,
  `add_groups`/`scale_groups`/`add_group_masked`, `is_bitsliced`. `number`/range
  re-expressed in terms of groups. Packed is the degenerate case
  `limbs_per_group == 1`.
- **`field/bitslice.rs`** (new) — const-K kernels (`1..=16`, heap fallback
  beyond) for add / masked-add / scale, plus hand-written F3 (6-gate) and F5
  flat circuits and the generic ripple-carry + conditional-subtract-p / 
  double-and-add scalar mult.
- **`Fp`** delegates `add_groups`/`scale_groups`/`add_group_masked` to the
  bit-slice kernels; **`SmallFq`** uses the same bit-sliced storage with the
  default element-wise (Zech-log) ops.
- **Vector/slice/matrix paths** route entry access and arithmetic through
  `gather`/`scatter`/`add_groups`. Slice addition handles equal nonzero lane
  offsets via a masked group path and arbitrary offsets via a general
  per-plane shifted add (this replaced the old `add_shift_*` methods).
- **Serde byte format** changes with the layout; golden files in
  `tests/serde_format.rs` regenerated.

## Testing

- Existing `field_tests!` and the layout-agnostic vector proptests pass against
  the bit-sliced layout.
- End-to-end `tests/row_reduce.rs` and the wider workspace resolution tests pass.
- `cargo +nightly fmt`, `cargo doc` (`-D warnings`), and clippy are clean.

## Summary by CodeRabbit

* **New Features**
  * Added bit-sliced prime-field storage with optimized bulk vector/matrix arithmetic, including masked lane updates for modular addition and modular scaling.
  * Extended vector iteration and slice element access so entry reads/writes work correctly in bit-sliced layouts.

* **Bug Fixes**
  * Fixed matrix conversion for bit-sliced primes by switching to per-entry set/get behavior.
  * Corrected group-boundary padding and improved mutable slice operations and arithmetic paths in bit-sliced mode.

* **Tests**
  * Updated serde golden JSON expectations for prime-field vectors to reflect the new limb layout.


ext/crates/fp/src/field/bitslice.rs
```diff
@@ -0,0 +1,533 @@
...
+    (u32::BITS - (p - 1).leading_zeros()) as usize
+}
+
+/// The bits of `p` as full-width lane masks: `out[j]` is all-ones iff bit `j` of `p` is set,
+/// for `j` in `0..=k`.
+fn p_masks(p: u32, k: usize) -> [Limb; BITS_PER_LIMB + 1] {
+    let mut masks = [0; BITS_PER_LIMB + 1];
+    for (j, m) in masks.iter_mut().enumerate().take(k + 1) {
+        *m = if (p >> j) & 1 == 1 { !0 } else { 0 };
+    }
```
hoodmane: Feels to me like this would be clearer to write as a loop:
```rs
for i in 0..min(k+1, BITS_PER_LIMB + 1) {
   if (p >> i) & 1 {
      masks[i] = !0;
   }
}
```

JoeyBF: Done in 8e81da3 — rewrote it with the explicit `if ... { = !0 }` instead of the ternary. I kept `masks.iter_mut().enumerate()` rather than a bare `for i in 0..n { masks[i] = ... }`, because the latter trips clippy's `needless_range_loop` and CI runs clippy with `-D warnings`. Same shape otherwise.

---

ext/crates/fp/src/field/bitslice.rs
```diff
@@ -0,0 +1,533 @@
...
+    masks
+}
+
+/// The full-width lane mask for bit `j` of `p`: all-ones if set, zero otherwise.
+#[inline(always)]
+fn pmask(p: u32, j: usize) -> Limb {
+    0u64.wrapping_sub(((p >> j) & 1) as Limb)
```
hoodmane: What's the deal with casting the argument `as Limb`?

JoeyBF: `(p >> j) & 1` is a `u32` (since `p: u32`), and the function returns a `Limb` (`u64`), so the bit has to be widened before the negate-to-mask trick (`1 -> !0`, `0 -> 0`). I've simplified it in 4c67bbc to `Limb::from((p >> j) & 1).wrapping_neg()`, which drops the `as` cast in favor of the lossless `From` and reads more directly.

---

PR #253, opened 2026-06-30
Ignore format commits for git blame

JoeyBF:
## Summary by CodeRabbit

* **Chores**
  * Added a Git blame ignore list to keep selected revision history out of blame results, making code history easier to read.


hoodmane:
Thanks!

PR #258, opened 2026-07-03
Fix Massey product panic on proper multiplication kernels (#116)

### LINKED ISSUE #116: Failure to compute Massey products <h3^2, d0, ->
If I'm not mistaken, the product h3^2 * d0 is zero (the target bidegree is empty). However, trying to compute the Massey products gives me the following error:

```sh
$ cargo run --example massey -- S_2 "" "" "" 14 2 "[1]" 14 4 "[1]"
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/examples/massey S_2 '' '' '' 14 2 '[1]' 14 4 '[1]'`
Module (default: S_2): S_2
Module save directory (optional):
Max n (default: 30):
Max s (default: 7):

Computing Massey products <a, b, ->

Enter a:
n of Ext class a: 14
s of Ext class a: 2
Input Ext class a: [1]

Enter b:
n of Ext class b: 14
s of Ext class b: 4
Input Ext class b: [1]
thread 'main' panicked at 'Failed to lift at (target_s, target_t) = (1, 3)', /home/joey/Documents/sseq/ext/src/chain_complex/chain_homotopy.rs:260:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

JoeyBF:
Fixes #116.

## Problem

`ExtAlgebra::massey_iter_c` and the single-shot `ExtAlgebra::massey` panic with `Failed to lift` (or, in release, the `apply_quasi_inverse` assertion) whenever the kernel of "multiply by `b`" at a third-factor bidegree is a *proper* subspace. The issue's reproduction still triggers it:

```
cargo run --example massey -- S_2 "" "" "" 14 2 "[1]" 14 4 "[1]"
```

(i.e. `a = h3²`, `b = d0`), panicking at `ext/src/chain_complex/chain_homotopy.rs:247`.

## Root cause

The triple Massey product `<a, b, c>` is defined only when `a·b = 0` and `b·c = 0`. `massey_at` built the null-homotopy of the composite `f_b ∘ f_genᵢ` **per generator `genᵢ`** of the third-factor bidegree, stored the per-generator bracket values in an `answers` matrix, and only *afterwards* computed the kernel of `·b`, combining `answers` linearly over the kernel rows.

But the null-homotopy of `f_b ∘ f_genᵢ` (multiplication by `b·genᵢ`) exists only when `b·genᵢ = 0` in Ext. When some generator is not individually killed by `b` (the kernel is a proper subspace), `ChainHomotopy::extend` cannot lift and panics. The per-generator "answers matrix, then combine over the kernel" design is fundamentally invalid for proper kernels: the bracket is only defined on the kernel and can only be realised by building the homotopy for the *actual* kernel class. This only surfaced for a first factor of filtration `s ≥ 2` (the cycle assertion at the homotopy's top step is gated on `target.s() > 1 == a.s() > 1`), which is why the existing tests — all using `h0` — never hit it.

## Fix

Restructure both entry points to realise the actual kernel class via `ResolutionHomomorphism::from_class` and build a single valid null-homotopy, mirroring the already-correct `massey_iter_a`:

- **`massey_kernel`** — the valid third factors (kernel of `·b`), computed from the product maps alone (no homotopy).
- **`massey_bracket_of`** — realises one kernel class and reads off its bracket.
- **`massey`** now verifies `b·c = 0` up front and never attempts to lift an invalid third factor.

`massey_at`/`MasseyComputeDatum` are removed; `massey_iter_a` and the shared `massey_result`/`massey_representative`/`massey_indeterminacy` helpers are unchanged.

This is not more expensive: `massey_iter_c` now builds one null-homotopy per kernel basis element (`dim(ker) ≤ num_gens`) instead of one per generator.

## Testing

- The issue's reproduction command now completes cleanly (exit 0) instead of panicking.
- `cargo clippy -p ext --lib` is clean; all `ext` library tests pass.
- New regression tests (both fit the `(6,5)` resolution already used by the existing tests):
  - `<h1², h0, h0>` returns `None` without panicking (first factor `s = 2`, exercising the single-shot path).
  - `massey_iter_c(h1², h0)` completes and agrees with the independent `massey_iter_a(h0, h1²)`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CAgVE4H22yxYTjSrH3r1iJ)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Massey product calculations by switching to a kernel-first approach for computing the valid third-factor.
  * Added stricter input validation so invalid third-factor conditions now return no result without panicking.
  * Fixed iteration over Massey products so it no longer stops early on problematic cases.
* **Tests**
  * Added regression coverage for edge cases (including higher filtration and normalization) to ensure consistent results.

### DIFF
--- a/ext/src/ext_algebra/massey.rs
+++ b/ext/src/ext_algebra/massey.rs
@@ -50,12 +50,6 @@ impl MasseyResult {
     }
 }
 
-struct MasseyComputeDatum {
-    answers: Matrix,
-    kernel: Subspace,
-    tot: Bidegree,
-}
-
 impl<CC> ExtAlgebra<CC>
 where
     CC: FreeChainComplex + AugmentedChainComplex,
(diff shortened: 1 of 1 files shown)


ext/src/ext_algebra/massey.rs
```diff
@@ -85,90 +79,109 @@ where
...
+    /// uncomputed, so callers never mistake an uncomputed product for a zero one; a computed but
+    /// empty product bidegree correctly yields the full space.
+    fn massey_kernel(&self, b: &BidegreeElement, c_deg: Bidegree) -> Option<Subspace> {
+        let p = self.prime();
+        let resolution = self.resolution();
+
+        let prod_deg = c_deg + b.degree();
+        if !resolution.has_computed_bidegree(prod_deg) {
+            return None;
+        }
```
hoodmane, line 96: We could also call `compute_through_bidegree`.

ext/src/ext_algebra/massey.rs
```diff
@@ -85,90 +79,109 @@ where
...
+    }
+
+    /// The bracket $\langle a, b, c\rangle$ for a single third factor `c`, which the caller must
+    /// have checked lies in the kernel of multiplication by `b` (so that `b · c = 0` and the
+    /// null-homotopy exists); otherwise the lift in [`ChainHomotopy::extend`] cannot complete.
+    ///
+    /// Unlike the removed per-generator scheme, this realises the *actual* class `c` (a linear
+    /// combination) via [`ResolutionHomomorphism::from_class`] and builds a single valid
+    /// null-homotopy, matching the approach of [`massey_iter_a`](Self::massey_iter_a). Returns
+    /// `None` when the bracket bidegree `c.degree() + shift` is uncomputed or empty.
```
hoodmane: > uncomputed or empty.

If it's empty the answer is 0 right? It's not some special case it's just the only element of the zero vector space. And again maybe we should just compute_through_bidegree as part of the function?

ext/src/ext_algebra/massey.rs
JoeyBF: I kept the `None` here for consistency with the rest of the Ext API: `multiply_into`/`try_multiply` and `massey_bracket_of` all return `None` on an uncomputed bidegree rather than resolving as a side effect, and the `massey` example resolves up front via `compute_through_stem`. Auto-computing only here would also be half a fix unless `massey_bracket_of` did the same for `tot`, and inside `massey_iter_c` it would silently extend the resolution mid-sweep. Happy to switch to on-demand `compute_through_bidegree` if you'd rather make that a deliberate API choice across the product/Massey helpers.

---

ext/src/ext_algebra/massey.rs
JoeyBF: Agreed on the empty case — fixed in b9a7c51: a computed-but-empty `tot` now returns `Some` with a zero representative (the unique element of the zero group), and `None` is reserved for a genuinely uncomputed `tot`. This only changes single-shot `massey`; `massey_iter_c` already filters brackets containing zero. On `compute_through_bidegree`, see my reply on the line-96 thread — I've kept the uncomputed case as `None` for consistency with `multiply_into`/`massey_bracket_of`, but happy to revisit if you'd like it done on demand.

---

hoodmane (approved):

PR #263, opened 2026-07-10
Genericize the algebra crate and resolution engine over the ground ring

JoeyBF:
## Summary

The `algebra` crate and the `ext` resolution engine hardcoded the coefficient
field 𝔽ₚ throughout. This PR makes them generic over an arbitrary graded
**ground ring**, introducing a small trait tower so the same engine can resolve
over rings other than a field — while keeping the classical 𝔽ₚ path
behavior-preserving and bit-identical.

No new coefficient ring beyond 𝔽ₚ is added here; this is the abstraction layer
that a follow-up (C-motivic Ext over 𝔽₂[τ]) builds on.

## Design

A coefficient ring is folded into the existing algebra hierarchy as a sub-trait,
and the "solving" linear algebra is split off into its own trait so it can be
required only where a resolution actually needs it:

- **`Ring: Algebra`** (`algebra/base_ring.rs`) — a coefficient ring *is* an
  algebra, plus a scalar type (`Ring::Element`), a representation of finite free
  modules over it (`Ring::Vector` and slices), and by-value arithmetic. This is
  enough to *define* an algebra and act on its modules.
- **`Solvable: Ring`** (`linear_algebra/`, new module) — the harder linear
  algebra a free resolution needs: images, kernels, quasi-inverses, and the
  resolution-step machinery. It is bounded only at `ModuleHomomorphism` /
  `FreeModuleHomomorphism` / `ChainComplex`, so an algebra can be *defined* over
  a ring whose solving linear algebra isn't implemented yet.
- **`Algebra`** gains `type BaseRing: Ring` + `base_ring()`, and a
  `module_at(t) -> GradedPiece` accessor exposing each graded piece as a
  `Module<Algebra = BaseRing>`.

Coefficients and vectors are threaded generically via the aliases `Scalar<A>`,
`VectorOf<A>`, `BaseSliceOf<A>`, `SubmoduleOf<A>`, `QuasiInverseOf<A>` (all
projected through `A::BaseRing`). The module actions, free modules, module
homomorphisms, chain complexes, and resolution steps are rewritten against these
aliases instead of concrete `fp` types.

`Field` (= 𝔽ₚ) is the reference implementer of both `Ring` (`Element = u32`,
`Vector = FpVector`) and `Solvable` (every operation forwards to `fp`).

## Compatibility & validation

- **Behavior-preserving for the classical path.** With `BaseRing = Field`
  everything monomorphizes to the same `fp` code as before. All **69 resolution
  benchmarks are bit-identical**, and the full test suite passes.
- `cargo fmt --check`, `clippy` (both `--no-default-features` and `--all-targets`
  profiles), and `cargo doc` under `-D warnings` are all green.
- The change is additive: no public classical API is removed; existing
  `Algebra<BaseRing = Field>` code is unaffected.

## Out of scope

- No motivic / non-field coefficient ring is introduced here. The 𝔽₂[τ] ground
  ring, its solving linear algebra, and the C-motivic Steenrod algebra are a
  follow-up that depends on this PR.

## Summary by CodeRabbit

* **New Features**
  * Added generic base-ring support across algebras, modules, module homomorphisms, and chain complexes.
  * Introduced ring-aware linear-algebra abstractions for vectors/slices, kernels/images, quasi-inverses, and resolution stages.
  * Exposed graded-piece access for supported algebras (including classical \(\mathbb{F}_p\) pieces).

* **Improvements**
  * Updated algebraic operations to use the appropriate scalar behavior for each base ring.
  * Resolution and homotopy-related computations now leverage the new linear-algebra workflow, with improved internal consistency and tests for graded pieces/linear algebra.

### DIFF
--- a/ext/crates/algebra/src/algebra/adem_algebra.rs
+++ b/ext/crates/algebra/src/algebra/adem_algebra.rs
@@ -19,8 +19,9 @@ use crate::{
     algebra::{
         Algebra, Bialgebra, Field, GeneratedAlgebra, UnstableAlgebra,
         combinatorics::{self, MAX_XI_TAU},
+        field::classical_graded_piece,
     },
-    module::{FreeModule, Module as _},
+    module::FreeModule,
 };
 
 /// An Adem basis element for the Steenrod algebra.
@@ -171,10 +172,7 @@ impl Algebra for AdemAlgebra {
     }
 
     fn module_at(&self, t: i32) -> FreeModule<Field> {
-        let piece = FreeModule::new(std::sync::Arc::new(self.base_ring()), String::new(), 0);
-        piece.add_generators(0, self.dimension(t), None);
-        piece.compute_basis(0);
-        piece
+        classical_graded_piece(self.base_ring(), self.dimension(t))
     }
 
     fn prefix(&self) -> &str {
--- a/ext/crates/algebra/src/algebra/adem_algebra.rs
+++ b/ext/crates/algebra/src/algebra/adem_algebra.rs
@@ -164,12 +164,11 @@ impl fmt::Display for AdemAlgebra {
 
 impl Algebra for AdemAlgebra {
     type BaseRing = Field;
+    type GradedPiece = FreeModule<Field>;
 
     fn base_ring(&self) -> Field {
         Field::new(self.prime())
     }
-
-    type GradedPiece = FreeModule<Field>;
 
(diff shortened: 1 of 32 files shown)


JoeyBF:
We're exploring a way to handle motivic computations that doesn't depend on more general ground rings. We'll revisit if we need to.

PR #268, opened 2026-07-13
Make the Algebra and Module traits multigrading-capable

### LINKED ISSUE #259: Split `ExtAlgebra` into a ring `ExtAlgebra` (over k) and an `ExtModule` (over M)
## Motivation

`ExtAlgebra<CC>` currently conflates two different objects:

- **the ring** `Ext(k, k)` — genuinely an algebra;
- **the module** `Ext(M, k)` — a *module* over that ring, not an algebra (the struct's own docs note "when `M == k` this is the algebra `Ext(k, k)` itself").

One struct plays both roles, carrying `resolution` (of `M`), `unit` (of `k`), an `is_unit` flag, and a `products` cache keyed on generators of the **resolution/`M` side** (`resolution → unit` maps).

The cache therefore has no home for a purely **ring-side** operation: multiplication by a class `b ∈ Ext(k, k)`. This surfaced while fixing #116 (see #258): `massey_b_hom` has to build the multiply-by-`b` chain map (`unit → unit`) fresh via `ResolutionHomomorphism::from_class` every call, because when `M ≠ k` that map is not what `products` stores. It only coincides with a cached map in the `M == k` case, and even then only for single generators (the cache is per-generator; class products are assembled at the `hom_k` level, not as a single chain map).

## Proposed structure

- **`ExtAlgebra`** — owns the resolution of `k` and the **ring-product cache** (products among `Ext(k, k)` generators, `unit → unit`, built and extended once). This becomes the single source of the multiply-by-`b` maps that `massey_b_hom` needs.
- **`ExtModule`** — owns the resolution of `M`, a shared `Arc<ExtAlgebra>` (a shared handle, so all modules over the same `k` share **one** ring cache — not a per-module copy, or the caching wouldn't actually be shared), and the **module-action cache** (`M`'s Ext-generators acted on by ring elements, `resolution → unit`).
- `M == k` collapses to "a module over itself," retiring the `is_unit` special-casing and `without_unit`.

## Benefits

- The ring multiplication (mult-by-`b`) is cached once in `ExtAlgebra` and reused across every module and every product/Massey computation.
- Correct semantics and naming: algebra vs. module.
- The Massey code separates cleanly: `massey_b_hom` → cached map from the `ExtAlgebra`; `massey_kernel` → module-action cache; `massey_bracket_of` composes the two. The `b` (ring-side) vs `c` (module-side) asymmetry in `massey_bracket_of` stops being a smell.

## Scope / notes

This touches all of `ext/src/ext_algebra/` (`mod.rs`, `massey.rs`, `secondary.rs` + `SecondaryExtAlgebra`), every example/consumer that constructs an `ExtAlgebra`, and possibly the web GUI. It is intentionally out of scope for the #116 fix (#258), which stays a focused bug fix. Related: #184 (cohomology abstraction layer).

<sub>Filed from discussion on #258.</sub>

JoeyBF:
## What & why

Generalize the `Algebra`/`Module` trait family to be **multigrading-capable**, so a bigraded object (like `Ext`) can implement them directly. This is groundwork for #259 (splitting `ExtAlgebra` into a ring `Ext(k, k)` and a module `Ext(M, k)` that *are* an `Algebra` and a `Module`); on its own it is a **pure generalization with zero behavior change** for all existing single-graded code.

The grading types already exist in the `sseq` crate (`MultiDegree<N>`, with `Bidegree = MultiDegree<2>`); this PR just makes the traits consume them.

## How

- **`trait Algebra<const N: usize = 1>`** and **`trait Module<const N: usize = 1>`**. The defaulted const generic means every existing `A: Algebra` / `M: Module` bound and the `Arc<dyn Module>` alias keep working unchanged — only future bigraded code writes `<2>`.
- **Degree *inputs* take `impl Into<MultiDegree<N>>`.** With `From<i32> for MultiDegree<1>`, every existing `i32` call site (~350 of them) compiles unchanged; the method `.into()`s at the top. **Degree *returns* stay `i32`** (the distinguished filtration direction), so the resolution engine's degree arithmetic is entirely untouched.
- **Object safety.** `Module` is used as `dyn Module`, and `impl Into<…>` arguments are not object-safe. So the object-safe core methods take a concrete `MultiDegree<N>` and are suffixed `_multi` (`dimension_multi`, `act_on_basis_multi`, …); a blanket **`ModuleExt`** extension trait re-exposes the ergonomic names (`dimension`, `act_on_basis`, …) taking `impl Into<MultiDegree<N>>`, and works even on `dyn Module`. Downstream code just gains `use algebra::module::ModuleExt`.
- **enum_dispatch.** It cannot dispatch a const-generic trait parameter, so `Algebra` is removed from `SteenrodAlgebra`'s `#[enum_dispatch(...)]` list and hand-dispatched via the existing `dispatch_steenrod!` macro (the same pattern `PairAlgebra` already uses for its associated type). `UnstableAlgebra`/`GeneratedAlgebra`/`MuAlgebra` stay singly graded and keep `#[enum_dispatch]`.

Only the ~3 concrete algebras (Milnor, Adem, Field) and ~8 module implementors gain `impl Into` + a one-line `.into()` per method; the bulk of the diff is those mechanical signature updates plus `ModuleExt` imports at call sites.

## Testing

- `cargo test --features concurrent --lib --tests --workspace`, `cargo test --examples`, and doctests: green (the lone `test_tempdir_lock` failure is a pre-existing sandbox/permissions artifact, unrelated — it fails on `master` too).
- Nightly `cargo fmt --check`, the clippy matrix (`--no-default-features` / `--all-targets`, `-D warnings`), and rustdoc (`-D warnings`, private items, all features): clean.
- The single-graded code path is byte-for-byte identical; this PR adds no new behavior.

## Follow-ups

The actual #259 split (`ExtAlgebra` ring / `ExtModule` module implementing these traits) builds on this and will come as a separate PR, followed by the secondary-layer split. This PR intentionally does **not** close #259.

https://claude.ai/code/session_01Nk9J57zb6GpvZxrduhGSdY

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nk9J57zb6GpvZxrduhGSdY)_

## Summary by CodeRabbit

- **New Features**
  - Added support for multi-dimensional grading across algebra and module operations.
  - APIs now accept flexible degree representations for dimensions, actions, basis generation, and element formatting.
  - Added convenient helper methods for common module operations.
  - Extended support for concurrent processing and odd-prime calculations.
- **Compatibility**
  - Existing singly graded workflows continue to work with the updated interfaces.

### DIFF
--- a/ext/crates/algebra/src/algebra/algebra_trait.rs
+++ b/ext/crates/algebra/src/algebra/algebra_trait.rs
@@ -6,6 +6,7 @@ use fp::{
     vector::{FpSlice, FpSliceMut},
 };
 use itertools::Itertools;
+use sseq::coordinates::MultiDegree;
 
 /// A graded algebra over $\mathbb{F}_p$.
 ///
@@ -20,8 +21,16 @@ use itertools::Itertools;
 /// this function before performing other operations at that degree.
 ///
 /// Algebras may have a distinguished set of generators; see [`GeneratedAlgebra`].
-#[enum_dispatch]
-pub trait Algebra: std::fmt::Display + Send + Sync + 'static {
+///
+/// # Grading
+/// The trait is generic over the number of gradings `N` (defaulting to `1`, the singly-graded
+/// case). Degrees are [`MultiDegree<N>`]; every degree *input* is taken as `impl Into<MultiDegree<N>>`
+/// so callers in the singly-graded world keep passing bare `i32`s (via `From<i32> for
+/// MultiDegree<1>`). `MultiDegree<2>` is a bidegree, used by `Ext` (see `ext::ext_algebra`).
+///
+/// `Algebra` cannot use `#[enum_dispatch]` because it is now generic; `SteenrodAlgebra` dispatches
+/// it by hand (see `dispatch_steenrod!`).
+pub trait Algebra<const N: usize = 1>: std::fmt::Display + Send + Sync + 'static {
     /// A name for the algebra to use in serialization operations. This defaults to "" for algebras
     /// that don't care about this problem.
     fn prefix(&self) -> &str {
@@ -49,10 +58,10 @@ pub trait Algebra: std::fmt::Display + Send + Sync + 'static {
     ///
     /// This function must be idempotent and cheap to call again with the
     /// same argument.
-    fn compute_basis(&self, degree: i32);
+    fn compute_basis(&self, degree: impl Into<MultiDegree<N>>);
 
     /// Returns the dimension of the algebra in degree `degree`.
-    fn dimension(&self, degree: i32) -> usize;
+    fn dimension(&self, degree: impl Into<MultiDegree<N>>) -> usize;
 
     /// Computes the product `r * s` of two basis elements, and adds the
     /// result to `result`.
@@ -62,9 +71,9 @@ pub trait Algebra: std::fmt::Display + Send + Sync + 'static {
         &self,
         result: FpSliceMut,
         coeff: u32,
-        r_degree: i32,
+        r_degree: impl Into<MultiDegree<N>>,
         r_idx: usize,
-        s_degree: i32,
+        s_degree: impl Into<MultiDegree<N>>,
         s_idx: usize,
     );
 
@@ -76,12 +85,14 @@ pub trait Algebra: std::fmt::Display + Send + Sync + 'static {
         &self,
         mut result: FpSliceMut,
         coeff: u32,
-        r_degree: i32,
+        r_degree: impl Into<MultiDegree<N>>,
         r_idx: usize,
-        s_degree: i32,
+        s_degree: impl Into<MultiDegree<N>>,
         s: FpSlice,
     ) {
         let p = self.prime();
+        let r_degree = r_degree.into();
+        let s_degree = s_degree.into();
         for (i, v) in s.iter_nonzero() {
             self.multiply_basis_elements(
                 result.copy(),
@@ -102,12 +113,14 @@ pub trait Algebra: std::fmt::Display + Send + Sync + 'static {
         &self,
         mut result: FpSliceMut,
         coeff: u32,
-        r_degree: i32,
+        r_degree: impl Into<MultiDegree<N>>,
         r: FpSlice,
-        s_degree: i32,
+        s_degree: impl Into<MultiDegree<N>>,
         s_idx: usize,
     ) {
         let p = self.prime();
+        let r_degree = r_degree.into();
+        let s_degree = s_degree.into();
         for (i, v) in r.iter_nonzero() {
             self.multiply_basis_elements(
                 result.copy(),
@@ -128,12 +141,14 @@ pub trait Algebra: std::fmt::Display + Send + Sync + 'static {
         &self,
         mut result: FpSliceMut,
         coeff: u32,
-        r_degree: i32,
+        r_degree: impl Into<MultiDegree<N>>,
         r: FpSlice,
-        s_degree: i32,
+        s_degree: impl Into<MultiDegree<N>>,
         s: FpSlice,
     ) {
         let p = self.prime();
+        let r_degree = r_degree.into();
+        let s_degree = s_degree.into();
         for (i, v) in s.iter_nonzero() {
             self.multiply_element_by_basis_element(
                 result.copy(),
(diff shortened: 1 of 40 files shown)


JoeyBF:
Pushed `a6b6947` addressing the review + a CI fix:

- **`is_unit` for `N > 1`** — applied the guard (returns `false` for multigraded modules, overridable), since `min_degree`/`max_degree` are the single `i32` filtration and can't characterize the unit when degrees share a `t`. `N = 1` behavior is unchanged.
- **`sseq_gui` build fix** — `web_ext/sseq_gui/src/actions.rs` needed `use algebra::module::ModuleExt` for the renamed `dimension`/`basis_element_to_string` methods. This is what broke the `lint` and `webserver` jobs; a downstream consumer of the rename I'd missed. (The `sseq.rs` `dimension` calls are on a different type and are unaffected.)

I **did not** add the suggested output validation to `try_act_multi` / `try_act_on_basis_multi`. Those methods are unchanged from `master` (`a3a63cb`) apart from the `_multi` rename, so adding it is out of scope for this zero-behavior-change PR — and a strict `result`-length check would contradict the documented contract that "the length of `input` need not be equal to the dimension of the module" (the flexibility resolution-to-a-stem relies on). Happy to tackle output validation as a separate change if you'd like it.

Context: this PR is deliberately just the multigrading groundwork; the `ExtAlgebra`/`ExtModule` split for #259 (which CodeRabbit's "Linked Issues" check flags as missing) is the follow-up PR that builds on this one.

---

ext/crates/algebra/src/algebra/algebra_trait.rs
```diff
@@ -20,8 +21,16 @@ use itertools::Itertools;
...
-pub trait Algebra: std::fmt::Display + Send + Sync + 'static {
+///
+/// # Grading
+/// The trait is generic over the number of gradings `N` (defaulting to `1`, the singly-graded
+/// case). Degrees are [`MultiDegree<N>`]; every degree *input* is taken as `impl Into<MultiDegree<N>>`
+/// so callers in the singly-graded world keep passing bare `i32`s (via `From<i32> for
+/// MultiDegree<1>`). `MultiDegree<2>` is a bidegree, used by `Ext` (see `ext::ext_algebra`).
+///
+/// `Algebra` cannot use `#[enum_dispatch]` because it is now generic; `SteenrodAlgebra` dispatches
+/// it by hand (see `dispatch_steenrod!`).
```
hoodmane:
```suggestion
/// The trait is generic over the number of gradings `N`. Degrees are [`MultiDegree<N>`].
/// Callers in the singly-graded world can keep passing bare `i32`s.
///
/// `Algebra` cannot use `#[enum_dispatch]` because it is generic; `dispatch_steenrod!` does dispatch 
/// by hand.
```

ext/crates/algebra/src/algebra/steenrod_algebra.rs
```diff
@@ -53,7 +54,10 @@ impl std::str::FromStr for AlgebraType {
 }
 
 #[allow(clippy::large_enum_variant)]
-#[enum_dispatch::enum_dispatch(Algebra, Bialgebra, GeneratedAlgebra, UnstableAlgebra)]
+// `Algebra` is now generic (`Algebra<const N>`), which `enum_dispatch` cannot handle, so it is
+// dispatched by hand below via `dispatch_steenrod!`. The remaining (non-generic) traits still use
+// `enum_dispatch`.
```
hoodmane:
```suggestion
// `Algebra` is generic, which `enum_dispatch` cannot handle, so it is dispatched by hand
// below via `dispatch_steenrod!`.
```

ext/crates/sseq/src/coordinates/degree.rs
```diff
@@ -66,6 +66,21 @@ impl<const N: usize> From<[i32; N]> for MultiDegree<N> {
     }
 }
 
+/// A single-graded degree is just an `i32`. This lets every `i32` degree in the singly-graded
+/// (`N == 1`) world coerce into a `MultiDegree<1>`, so callers can keep passing bare `i32`s to the
+/// multigraded trait methods (which take `impl Into<MultiDegree<N>>`).
```
hoodmane, line 71:
```suggestion
```

hoodmane (approved):

PR #269, opened 2026-07-15
Migrate the algebra benchmark suite to criterion

JoeyBF:
Splits the benchmark infrastructure out of the Nassau GPU work (#264) so it can land on its own.

## What

- Replaces the `bencher` harness with `criterion` (plus `pprof` for flamegraph output).
- Adds a general benchmark suite for the algebra crate:
  - `benches/common/mod.rs` — shared helpers for algebra-multiplication and module-action benchmarks, plus Steenrod-module loading.
  - `benches/milnor.rs` — the Milnor multiply, ported to criterion.
  - `benches/multiplication.rs` — element multiplication across Adem/Milnor.
  - `benches/module_action.rs` — module actions.
  - `benches/nassau_milnor.rs` — the Nassau-regime (p = 2) Milnor multiply.

## Notes

This is pure benchmark infrastructure — no source-code changes, no dependency on the GPU work. It builds against the current `master`.

## Test plan

- [x] `cargo check -p algebra`
- [x] `cargo bench --no-run -p algebra` — every bench target compiles (`milnor`, `multiplication`, `module_action`, `nassau_milnor`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01UPYvsLdEfitgCbAiPxxx3U)_

## Summary by CodeRabbit

* **Performance**
  * Added expanded benchmarking for algebra multiplication, module actions, and Nassau–Milnor multiplication.
  * Added coverage across supported primes, algebra representations, and predefined Steenrod modules.
  * Added flamegraph profiling and HTML benchmark reports for easier performance analysis.
  * Migrated existing Milnor benchmarks to the new benchmarking framework.

PR #270, opened 2026-07-15
Add hash-free seqno index tables to the Milnor algebra

JoeyBF:
Splits the GPU-agnostic `seqno` index out of the Nassau GPU work (#264) so it can land on its own. Second of three; follows the benchmark split (#269).

## What

A flat, arc-swapped hash-free index for the Milnor basis: an O(number of p-part entries) rank computed from a precomputed `g` table, with no hash lookup.

- `SeqnoTables` + `compute_seqno_tables` (idempotent, concurrency-safe via `ArcSwapOption::rcu`) + `seqno`, applicable at `p = 2` with a trivial profile and stable ordering.
- `SeqnoRanker`, which acquires the tables once for a batch of lookups instead of taking an `arc_swap` guard per call.
- `benches/seqno.rs`: an A/B of the table index against the basis hashmap across a range of degrees.

## Why it is worth having

Three reasons, in descending order of how settled they are.

**1. The GPU path requires it.** A kernel cannot carry a hashmap, so an arithmetic rank is the only way to turn a `p_part` into a basis index on-device, and the flat `g` table uploads directly. This is the blocking dependency for #271.

**2. It is dramatically smaller, at every degree.** The basis hashmap holds one entry per basis element and is built for every degree up to the max; the `g` table is shared across all degrees and grows only linearly in the degree:

| max degree | dim at that degree | basis hashmap, all degrees | `g` table |
|---|---|---|---|
| 200 | 15,499 | 19 MB | 16 KB |
| 300 | 87,977 | 144 MB | 24 KB |
| 350 | 178,195 | 329 MB | 27 KB |
| 400 | 335,566 | 688 MB | 31 KB |
| 450 | 596,191 | 1.3 GB | 35 KB |

Four orders of magnitude, and unlike the speed comparison it does not have a crossover — the table is smaller everywhere. This saving is *available* rather than banked: realising it means not building the hashmap at all in the applicable regime, which is the follow-up below, not this PR.

**3. Above a crossover it is also faster.** Per lookup, at `p = 2`:

| degree | dim | hashmap | seqno | |
|---|---|---|---|---|
| 32 | 47 | 7.09 ns | 12.39 ns | |
| 100 | 1,189 | 7.39 ns | 15.59 ns | |
| 180 | 10,155 | 8.53 ns | 17.18 ns | |
| 260 | 46,750 | 13.24 ns | 18.15 ns | |
| 300 | 87,977 | 15.26 ns | 18.07 ns | |
| 320 | 117,834 | 22.27 ns | **18.70 ns** | seqno wins |
| 370 | 231,354 | 46.76 ns | **18.98 ns** | 2.5x |
| 400 | 335,566 | 53.27 ns | **19.56 ns** | 2.7x |

`seqno` is flat — 12 ns to 20 ns while the dimension grows four orders of magnitude — while the hashmap degrades 7.5x as its per-degree map leaves cache. They cross between degree 300 and 320.

This is the behaviour Christian Nassau has reported since 1998, when the comparison was against binary search over a written-out basis. It survives the move to a hashmap keyed on the packed p-part (#280): that changed the constant factor, not the asymptotics.

## Scope: why this PR does not yet dispatch on it

`try_basis_element_to_index` still uses the hashmap unconditionally, and `compute_basis` does not build the tables. Two things have to be true before flipping that, and only one of them is:

- **Where the time goes.** Counting `basis_element_to_index` calls over a real Nassau resolution of `S_2`: 2.46 M calls at t = 60, 1.21 billion at t = 120 — but against ~426 ns of surrounding work per call, so the lookup is only ~1.7% of runtime. The whole prize is under 2%. The looked-up degree tracks about 0.65·t, so at t = 350 roughly 7% of lookups are past the crossover and the weighted average still favours the hashmap; whole-resolution crossover is near t ≈ 420.
- **The benchmark flatters the hashmap.** It hammers one degree's map in a tight loop with nothing competing. A real high-degree run has hundreds of MB of index maps live against a 33 MB L3, so in situ the hashmap is worse than the table above shows and the true crossover is lower than 320 — by how much, a microbenchmark cannot say.

So the dispatch wants an in-situ profile rather than these numbers, and the memory argument (2) may well decide it before the timing one does. That is a separate PR; this one lands the primitive, the benchmark, and the evidence.

## Notes

- `seqno` takes the degree from the caller, which always has it, rather than re-deriving it as `Σ rᵢ·ξᵢ`. With `SeqnoRanker` hoisting the guard, that is worth 2.2x-2.8x — an earlier version of this benchmark was measuring that self-inflicted overhead as though it were inherent.
- `compute_seqno_tables` asserts the degree bound `compute_basis` already enforces. A negative argument wrapped to a huge row count (CodeRabbit).

## Test plan

- [x] `cargo test -p algebra` — incl. `seqno_matches_enumeration_order`, which cross-checks the caller-supplied degree against the p-part in debug builds
- [x] `cargo bench -p algebra --bench seqno` — the timing table above
- [x] Call counts measured with a temporary counter on `basis_element_to_index` over `construct(("S_2", "milnor"))` resolutions; the counting atomic itself cost 0.9% (518.3 s vs 513.7 s at t = 120), so it does not distort the ratio
- [x] `just lint` (nightly fmt + `cargo hack clippy` feature-powerset, `-D warnings`)

---
_Generated by [Claude Code](https://claude.ai/code/session_01UPYvsLdEfitgCbAiPxxx3U)_

### DIFF
--- a/ext/crates/algebra/benches/seqno.rs
+++ b/ext/crates/algebra/benches/seqno.rs
@@ -1,0 +1,70 @@
+//! A/B benchmark: the hash-free table-based Milnor index ([`MilnorAlgebra::seqno`]) versus the
+//! `FxHashMap`-backed [`MilnorAlgebra::basis_element_to_index`], over the exact operation of
+//! turning a basis element back into its index.
+//!
+//! This is the lookup that a resolution performs on *every* output term of a multiply — Nassau's
+//! `S_2` @ p=2 run issues ~1.3B of them — so a lookup that is even a little cheaper is worth having,
+//! and a GPU kernel (which cannot carry a hashmap) *needs* the table-based form regardless.
+//!
+//! The `seqno` group here reads the flat [`arc_swap`]-backed table; compare its `basis_to_index/*`
+//! ids against the `hashmap/*` ids at the same degree to see which wins on the CPU.
+
+use std::hint::black_box;
+
+use algebra::{Algebra, MilnorAlgebra};
+use criterion::{Criterion, Throughput, criterion_group, criterion_main};
+use fp::prime::TWO;
+use pprof::criterion::{Output, PProfProfiler};
+
+/// Degrees to sweep. Chosen to bracket the range a real `S_2` resolution reaches, from the cheap
+/// low-dimensional end up to degrees whose basis has thousands of elements.
+const DEGREES: &[i32] = &[16, 24, 32, 40, 48, 56, 64];
+
+fn seqno(c: &mut Criterion) {
+    let algebra = MilnorAlgebra::new(TWO, false);
+    let max_degree = *DEGREES.iter().max().unwrap();
+    algebra.compute_basis(max_degree);
+    algebra.compute_seqno_tables(max_degree);
+
+    let mut g = c.benchmark_group("seqno");
+
+    for &degree in DEGREES {
+        let dim = algebra.dimension(degree);
+        if dim == 0 {
+            continue;
+        }
+        // Snapshot the basis so neither method pays to walk the algebra's storage during timing.
+        let basis: Vec<_> = (0..dim)
+            .map(|i| algebra.basis_element_from_index(degree, i).clone())
+            .collect();
+
+        g.throughput(Throughput::Elements(dim as u64));
+
+        g.bench_function(format!("hashmap/deg{degree}"), |b| {
+            b.iter(|| {
+                for elt in &basis {
+                    black_box(algebra.basis_element_to_index(elt));
+                }
+            });
+        });
+
+        g.bench_function(format!("basis_to_index/deg{degree}"), |b| {
+            b.iter(|| {
+                for elt in &basis {
+                    black_box(algebra.seqno(&elt.p_part));
+                }
+            });
+        });
+    }
+
+    g.finish();
+}
+
+criterion_group! {
+    name = benches;
+    config = Criterion::default()
+        .measurement_time(std::time::Duration::from_secs(3))
+        .with_profiler(PProfProfiler::new(100, Output::Flamegraph(None)));
+    targets = seqno
+}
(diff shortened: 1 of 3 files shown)


ext/crates/algebra/benches/seqno.rs
```diff
@@ -0,0 +1,71 @@
+//! A/B benchmark: the hash-free table index against the basis hashmap.
+
+use algebra::{Algebra, MilnorAlgebra};
+use criterion::{Criterion, Throughput, black_box, criterion_group, criterion_main};
+use fp::prime::TWO;
+
+/// Degrees to sample, spanning the point where the table index overtakes the hashmap (see
+/// `EXPERIMENTS.md`).
```
hoodmane, line 8:
```suggestion
/// Degrees to sample, spanning the point where the table index overtakes the hashmap.
```

ext/crates/algebra/benches/seqno.rs
```diff
@@ -0,0 +1,71 @@
+//! A/B benchmark: the hash-free table index against the basis hashmap.
+
+use algebra::{Algebra, MilnorAlgebra};
+use criterion::{Criterion, Throughput, black_box, criterion_group, criterion_main};
+use fp::prime::TWO;
+
+/// Degrees to sample, spanning the point where the table index overtakes the hashmap (see
```
hoodmane, line 7: Maybe this could use a bit of expansion here. We have two approaches: this and hashmap. Hashmap good in low degrees, seqno good in high degree.

hoodmane (approved):
Should we update the code to use it in dimension >= 320 before we merge it?

hoodmane:
Also you should clarify in the commit message that this code is actually useful. The current PR description makes it sound not very useful.

PR #272, opened 2026-07-17
Relax the dependency graph in Nassau's `compute_through_stem`

JoeyBF:
Implements the relaxation flagged by the long-standing comment in `nassau.rs`: computing `(s, t)` only needs `(s, t-1)` and `(s-1, t-1)`, not `(s-1, t)` and `(s, t-1)`. The comment noted that "having the dimensions of the modules change halfway through the computation is annoying to do correctly" — this handles that.

## What

- `compute_through_stem` now uses the relaxed wavefront `(s, t) <- (s, t-1), (s-1, t-1)` for `s >= 2`, keeping many `t`-diagonals (`n = t - s` fixed) in flight at once. Rows `s = 0, 1` stay strict — they're cheap, and `step0`/`step1` read their targets through full matrices.
- `step_resolution_with_subalgebra` reads its target `C_{s-1}` using only generators of degree `< t` (and `C_{s-2}` of degree `< t-1`). By minimality the differentials we lift land in the radical, so the degree-`t` generators — which `(s-1, t)` may be adding concurrently — contribute neither to the kernel we quotient by nor to those differentials. The read is thus race-free against those append-only writes, with no locking. New helpers: a generator-degree bound on `signature_mask`, plus `restricted_dimension` and `restricted_partial_matrix`.
- Differentials are stored in the restricted (radical) basis, and the quasi-inverse save files always take the existing "incomplete information" (`Magic::Fix`) path, which the secondary machinery already handles.
- `ModuleHomomorphism::apply_to_basis_element` now allows a result buffer shorter than the full target dimension (a prefix of the target basis); the matching truncated differential means `act` never writes past it. This is the only change outside `nassau.rs`.

## Notes

The relaxed graph shortens the critical path to `(S, T)` from ~`S + T` to ~`T`: a whole internal-degree column can compute in parallel once the previous one is committed. Nassau can afford this because each bidegree already computes the kernel it consumes locally (it reduces the *incoming* differential `d_{s-1}`), so relaxing shares nothing across bidegrees and duplicates no work.

### Remark: the classical resolution (`resolution.rs`)

The same relaxation is mathematically valid for `MuResolution` — `ker(d_{s-1})_t` is a column-`(t-1)` object there too (see the existing `get_kernel` comment) — but it is *not* a free win. The classical algorithm reduces the *outgoing* differential at each bidegree and gets `ker(d_{s-1})_t` as the byproduct of the single row reduction it does at `(s-1, t)` to place that row's generators, handing it forward through the kernel cache. That reuse is exactly the `(s-1, t) -> (s, t)` edge. Relaxing means decoupling "find kernel" (column-`(t-1)`) from "place generators" (which needs `(s-2, t)`) and scheduling/caching the kernel independently — the `get_kernel` stem-edge path is already a special case of this. Done carefully it costs no extra reductions, only holding the reduced matrix live a little longer (memory) plus the scheduling refactor; done naively (cache only the `Subspace`, recompute the matrix) it doubles the expensive reductions. Left out of this PR since p=2 workloads use Nassau in practice, but noted here for the record.

## Test plan

All with `--features concurrent` (the parallel path; `maybe_rayon` runs sequentially otherwise):

- [x] `cargo test -p ext --lib nassau::` — `test_restart_stem`, plus a new `test_stem_concurrent_secondary` (save-backed d2 cross-check against the standard resolution, exercising the `Magic::Fix` quasi-inverse path)
- [x] `cargo test -p ext --test milnor_vs_nassau` — existing `compare`, plus a new wide `compare_stem` cross-check against the standard resolution (S_2, C2 to n=40/s=30; Joker to n=30/s=20), run 12× at `RAYON_NUM_THREADS=16` for determinism
- [x] full `cargo test -p ext` — green except the pre-existing `save_load_resolution::test_tempdir_lock` (unrelated: it relies on read-only directory bits that root bypasses)
- [x] builds and tests pass in both the default and `concurrent` feature configurations

https://claude.ai/code/session_01MCUtWj6P6suSZqvATCdg6d

---
_Generated by [Claude Code](https://claude.ai/code/session_01MCUtWj6P6suSZqvATCdg6d)_

## Summary by CodeRabbit

* **New Features**
  * Added support for calculating partial basis-action results and matrices using caller-defined target dimensions.
  * Added dimension queries that account only for generators below a specified degree, improving consistency during ongoing computations.
  * Added bounded generator-degree filtering to Nassau resolution calculations.

* **Bug Fixes**
  * Improved Nassau computation scheduling and preservation of restricted calculation behavior.

* **Tests**
  * Added cross-checks comparing Nassau and standard resolutions across multiple degrees and boundary cases.

### DIFF
--- a/ext/crates/algebra/src/module/free_module.rs
+++ b/ext/crates/algebra/src/module/free_module.rs
@@ -316,8 +316,32 @@ impl<const U: bool, A: MuAlgebra<U>> MuFreeModule<U, A> {
         }
     }
 
+    /// The dimension in `degree` of the submodule spanned by the generators of degree strictly
+    /// less than `max_gen_degree`.
+    ///
+    /// Equivalently, the dimension in `degree` when we pretend the generators of degree
+    /// `>= max_gen_degree` do not exist. Use [`Module::dimension`] for the unrestricted count.
+    ///
+    /// This recomputes the offset by summing the operation dimensions over the generators below
+    /// `max_gen_degree`, rather than reading the stored one as [`Self::generator_offset`] does.
+    /// The stored offsets only exist for generators that have been added, so the offset one past
+    /// the last generator below `max_gen_degree` is not available until a generator of degree
+    /// `>= max_gen_degree` is added. Recomputing reads only generator counts of degree
+    /// `< max_gen_degree`, which lets a caller use this while another thread adds generators of
+    /// degree `max_gen_degree`.
+    pub fn dimension_from_gens_below(&self, degree: i32, max_gen_degree: i32) -> usize {
+        self.iter_gen_offsets([degree])
+            .take_while(|gen_data| gen_data.gen_deg < max_gen_degree)
+            .map(|gen_data| gen_data.end[0])
+            .last()
+            .unwrap_or(0)
+    }
+
     /// Given a generator `(gen_deg, gen_idx)`, find the first index in degree `degree` with
     /// elements from the generator.
+    ///
+    /// This reads the generator count in `gen_deg`. See [`Self::dimension_from_gens_below`] for a
+    /// variant that does not, at the cost of recomputing the offset.
     pub fn generator_offset(&self, degree: i32, gen_deg: i32, gen_idx: usize) -> usize {
         assert!(gen_deg >= self.min_degree);
         assert!(gen_idx < self.num_gens[gen_deg]);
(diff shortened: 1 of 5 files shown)


ext/src/nassau.rs
```diff
@@ -925,32 +1014,57 @@ impl<M: ZeroModule<Algebra = MilnorAlgebra>> Resolution<M> {
...
+        // largest committed `t` in row `s`, so it doubles as a "predecessor committed" test.
+        let ready = |s: i32, t: i32, progress: &[i32]| -> bool {
+            in_region(s, t)
+                && progress[s as usize] >= t - 1
+                && match s {
+                    0 => true,
+                    // Row 1's diagonal predecessor is (0, t). At the stem edge (t = max_n + 1) that
+                    // bidegree lies outside the computed region, so we treat it as satisfied.
+                    1 => t > max_n || progress[0] >= t,
+                    _ => progress[(s - 1) as usize] >= t - 1,
```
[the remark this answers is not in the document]
JoeyBF: This is correct as written, so I've kept it and added tests rather than the halo.

`(1, max_n + 1)` doesn't need `(0, max_n + 1)`: by the same augmentation-minimality that the `s ≥ 2` rows use, `extend_to_surjection` at `(0, t)` maps the degree-`t` generators of `C_0` to a complement of `im(f_0|_{<t})` in `M_t`, so they are linearly independent from the lower image and never lie in `ker(f_0)`. Hence `ker(f_0)_{max_n+1}` — the only thing `step1` needs there — lives entirely in the degree-`< max_n+1` part of `C_0`, which is fully committed. So `step1` computes the correct generators even though `(0, max_n + 1)` (which has stem `max_n + 1 > max_n`, i.e. outside the requested region) is never computed. There's also no race: nothing writes `C_0` at that degree.

The proposed halo would additionally compute `(0, max_n + 1)`, which is outside the requested stem.

Locked in with `compare_stem` cases at `max_n = 2^i - 1` (7 → `h_3` at `(1,8)`, 15 → `h_4` at `(1,16)`), which put a nonzero class exactly on this edge and match the standard resolution.

---

ext/crates/algebra/src/module/homomorphism/free_module_homomorphism.rs
```diff
@@ -162,6 +145,81 @@ where
...
+                .maybe_par_iter_mut()
+                .enumerate()
+                .for_each(|(i, row)| {
+                    self.apply_to_basis_element_restricted(row, 1, degree, inputs[i])
+                });
+        }
+        matrix
+    }
+
+    /// The body of the two `apply_to_basis_element` variants, with no length check on `result`.
```
hoodmane, line 193:
```suggestion
```

ext/crates/algebra/src/module/homomorphism/free_module_homomorphism.rs
```diff
@@ -162,6 +145,81 @@ where
...
+    ///
+    /// Unlike [`ModuleHomomorphism::get_partial_matrix`], which sizes the matrix from the target's
+    /// dimension, this takes the number of columns from the caller. Each row is filled by
+    /// [`Self::apply_to_basis_element_restricted`], so the same support requirement applies.
+    pub fn get_partial_matrix_restricted(
+        &self,
+        degree: i32,
+        inputs: &[usize],
+        target_dim: usize,
+    ) -> Matrix {
```
hoodmane, line 180: Maybe unify with `get_partial_matrix` and make `target_dim` into `Option<usize>`?

ext/crates/algebra/src/module/free_module.rs
```diff
@@ -316,8 +316,32 @@ impl<const U: bool, A: MuAlgebra<U>> MuFreeModule<U, A> {
         }
     }
 
+    /// The dimension in `degree` of the submodule spanned by the generators of degree strictly
+    /// less than `max_gen_degree`.
+    ///
+    /// Equivalently, the dimension in `degree` when we pretend the generators of degree
+    /// `>= max_gen_degree` do not exist. Use [`Module::dimension`] for the unrestricted count.
```
hoodmane, line 323:
```suggestion
```

ext/crates/algebra/src/module/free_module.rs
```diff
@@ -316,8 +316,32 @@ impl<const U: bool, A: MuAlgebra<U>> MuFreeModule<U, A> {
...
+    ///
+    /// This recomputes the offset by summing the operation dimensions over the generators below
+    /// `max_gen_degree`, rather than reading the stored one as [`Self::generator_offset`] does.
+    /// The stored offsets only exist for generators that have been added, so the offset one past
+    /// the last generator below `max_gen_degree` is not available until a generator of degree
+    /// `>= max_gen_degree` is added. Recomputing reads only generator counts of degree
+    /// `< max_gen_degree`, which lets a caller use this while another thread adds generators of
+    /// degree `max_gen_degree`.
```
hoodmane, line 331:
```suggestion
    /// This recomputes the offset by summing the operation dimensions over the generators below
    /// `max_gen_degree`, rather than reading the stored one as [`Self::generator_offset`] does.
```

ext/src/nassau.rs
```diff
@@ -956,32 +1006,55 @@ impl<M: ZeroModule<Algebra = MilnorAlgebra>> Resolution<M> {
...
     /// This function resolves up till a fixed stem instead of a fixed t.
+    ///
+    /// The dependency graph we use is the relaxed one: computing `(s, t)` only requires
+    /// `(s, t - 1)` and `(s - 1, t - 1)` (for `s >= 2`), rather than `(s - 1, t)` and `(s, t - 1)`;
+    /// see `step_resolution_with_subalgebra` for why that suffices. This lets `(s, t)` run
+    /// concurrently with `(s - 1, t)` and keeps many `t`-diagonals (`n = t - s` fixed) in flight at
+    /// once, which is where the parallelism comes from.
+    ///
+    /// The rows `s = 0` and `s = 1` are kept on the strict schedule: `step0` and `step1` read their
+    /// targets through full matrices, so they wait for `(s - 1, t)`.
```
hoodmane, line 1017:
```suggestion
```

hoodmane (approved):

PR #275, opened 2026-07-18
nassau: recompute quasi-inverses on demand, with bidegree-major batched lifting

JoeyBF:
## Motivation

On-disk quasi-inverses (QIs) are the sole blocker to larger nassau computations: stem 256 already exceeds 20 TB and stem 300 would blow past 40 TB. The QIs solve the lifting problem so downstream work (products, Massey products, the secondary/`d₂` structure) doesn't recompute them — but they are mathematically re-derivable from the (tiny) saved differentials. The catch is that re-deriving one QI costs about as much as computing it did originally, so recomputing *per lift* is far too slow.

This PR makes recompute-on-demand practical by **amortizing each QI recompute across every lift that needs it**: sweep output bidegrees, and at each one solve the shared quasi-inverse once for all the maps/homotopies passing through it.

## What's here

**1. Recompute-on-demand for nassau QIs** (`nassau.rs`)
Regenerate a QI on the fly from `differentials[s]` alone (plus the module bases and the deterministically-chosen subalgebra) instead of reading it from disk. It streams one signature at a time through the exact byte format `write_qi` produces, so `apply_quasi_inverse` reads a recomputed QI through the same code path as a saved one and peak memory matches what resolving that bidegree originally took — never the whole QI, which can reach hundreds of GB at record stems. **Opt-in**, no default behavior change: `EXT_NASSAU_NO_SAVE_QI` stops persisting QIs, `EXT_NASSAU_RECOMPUTE_QI` forces recompute. By default QIs are still saved and read.

**2. Bidegree-major batched lifting** (`resolution_homomorphism.rs`)
A `Liftable` trait (target/source-type-free: `prepare(b) -> Option<{inputs, finish}>`) and a `MultiLift` driver that extends many liftables sharing one target complex together, in bidegree-major order via `iter_s_t`, issuing **one** batched `apply_quasi_inverse` per output bidegree. With QIs recomputed on demand this keeps a whole many-map computation at ~1× the on-disk baseline instead of scaling with the number of maps. The single-map lift paths become behavior-preserving `prepare_step`/`finish_step` splits.

**3. Consumers wired to batch**
- **Products** — `ExtAlgebra::extend_all_products` builds the product map per generator and extends them all through one `MultiLift`.
- **Massey** — `massey_iter_c` now builds every `f_c` and null-homotopy up front and extends them in two batched passes (maps, then homotopies) instead of a fresh homotopy per third factor. `massey_iter_a` already reuses one homotopy, so it's unchanged.
- **Secondary** — the `SecondaryLift` homotopy solve gets the same `prepare`/`finish` split and a `SecondaryLiftable` wrapper + `batch_extend_secondary`; `SecondaryExtAlgebra::extend_all_secondary_products` batches all secondary products (`SecondaryChainHomotopy` / secondary Massey is left for a follow-up).

**4. A latent `MultiLift` fix**
`step_cell` skipped the finishers whenever the *combined* inputs across all liftables at a bidegree were empty. A zero-dimensional step lifts nothing but must still run its finish to register/extend the step, or a later read of that bidegree indexes an unfilled degree and panics. Finishers now always run (with an empty results slice when there's nothing to lift).

## Behavior & compatibility

- Default behavior is unchanged — QIs are still saved and read; recompute is opt-in.
- All single-map paths are behavior-preserving refactors (the `prepare`/`finish` splits), pinned by the existing tests.

## Testing

- `test_iter_a_matches_iter_c` / `test_iter_c_proper_kernel` diff batched `massey_iter_c` against the independent `massey_iter_a` reference.
- `batched_matches_native_d2` pins the secondary wrapper against native `compute_homotopies`; `batched_secondary_products_match` pins the multi-lift batched products against the per-lift path.
- Existing products, Massey, secondary, and save/load tests still pass. `just lint` (nightly rustfmt + clippy) is clean on the touched files, and docs build under `-D rustdoc::broken_intra_doc_links`.

Illustrative timing (`all_products` on S₂, QIs recomputed): batched **4.57 s** vs. per-map **17.33 s** for 516 product maps — the redundancy the batching removes.

## Relationship to #260 (ZarrV3)

Deliberately built on `master`, independent of the ZarrV3 save-format PR, so it can land first — it settles *what* we store (recompute-on-demand removes the QI files) before that PR finalizes *how*. Once this lands, #260 can drop its QI stream tier.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KiaVy4c9APWdmFeFoPCbdo)_

## Summary by CodeRabbit

* **New Features**
  * Added batch computation for all primary and secondary product maps.
  * Added controls to save or recompute previously calculated quasi-inverses.
  * Added an example benchmark for comparing product-map extension modes.
* **Performance**
  * Improved large-scale product and Massey product calculations by sharing work across multiple lifts.
* **Tests**
  * Added coverage confirming batched secondary products match individually computed results.

### DIFF
--- a/ext/src/nassau.rs
+++ b/ext/src/nassau.rs
@@ -15,7 +15,7 @@
 use std::{
     fmt::Display,
     io,
-    sync::{Arc, Mutex, mpsc},
+    sync::{Arc, LazyLock, Mutex, mpsc},
 };
 
 use algebra::{
@@ -388,6 +388,17 @@ enum Magic {
     Fix = -3,
 }
 
+/// Whether to persist quasi-inverses to disk during resolution. Disabled by
+/// `EXT_NASSAU_NO_SAVE_QI`, in which case only the differentials are written (the quasi-inverses are
+/// ~260-460x larger) and every downstream lift recomputes its quasi-inverse on demand.
+static SAVE_QI: LazyLock<bool> =
+    LazyLock::new(|| std::env::var_os("EXT_NASSAU_NO_SAVE_QI").is_none());
+
+/// Force `apply_quasi_inverse` to recompute rather than read a saved quasi-inverse, even when one
+/// exists on disk. Used to measure the recompute cost in isolation.
+static RECOMPUTE_QI: LazyLock<bool> =
+    LazyLock::new(|| std::env::var_os("EXT_NASSAU_RECOMPUTE_QI").is_some());
+
 /// A resolution of `S_2` using Nassau's algorithm.
 ///
 /// This aims to have an API similar to that of
@@ -598,7 +609,11 @@ impl<M: ZeroModule<Algebra = MilnorAlgebra>> Resolution<M> {
         let next = &self.modules[b.s() - 2];
         next.compute_basis(b.t());
 
-        let mut f = if let Some(dir) = self.save_dir().write() {
+        // Skip writing the quasi-inverse when `EXT_NASSAU_NO_SAVE_QI` is set; `apply_quasi_inverse`
+        // recomputes it on demand from the differential (see `RecomputeReader`).
+        let mut f = if *SAVE_QI
+            && let Some(dir) = self.save_dir().write()
+        {
             let mut f = self
                 .save_file(SaveKind::NassauQi, b - Bidegree::s_t(1, 0))
                 .create_file(dir.to_owned(), true);
@@ -1057,14 +1072,23 @@ impl<M: ZeroModule<Algebra = MilnorAlgebra>> ChainComplex for Resolution<M> {
         for<'a> &'a mut T: Into<FpSliceMut<'a>>,
         for<'a> &'a S: Into<FpSlice<'a>>,
     {
-        let mut f = if let Some(dir) = self.save_dir.read() {
-            if let Some(f) = self.save_file(SaveKind::NassauQi, b).open_file(dir.clone()) {
-                f
-            } else {
-                return false;
-            }
+        // Read the saved quasi-inverse unless recomputation is forced. Fall back to recomputation
+        // whenever no saved stream is available: no store, the qis were never persisted
+        // (`EXT_NASSAU_NO_SAVE_QI`), or the store has no qi for this bidegree. The last case
+        // legitimately happens at the top of the computed region — nassau writes qi(s, t) while
+        // computing (s + 1, t), so qi(max_s, t) is never saved even though lifting into it is
+        // well-defined. `RecomputeReader` regenerates the exact same byte stream from
+        // `differentials[b.s]`, so the loop below is unchanged.
+        let saved = if *RECOMPUTE_QI {
+            None
+        } else if let Some(dir) = self.save_dir.read() {
+            self.save_file(SaveKind::NassauQi, b).open_file(dir.clone())
         } else {
-            return false;
+            None
+        };
+        let mut f: Box<dyn io::Read> = match saved {
+            Some(f) => f,
+            None => Box::new(RecomputeReader::new(self, b)),
         };
 
         let p = self.prime();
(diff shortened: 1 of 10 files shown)


ext/src/nassau.rs
```diff
@@ -1224,6 +1246,148 @@ impl<M: ZeroModule<Algebra = MilnorAlgebra>> ChainComplex for Resolution<M> {
...
+/// Because the resolution is fully computed by the time a lift is requested, the recomputed
+/// quasi-inverse always uses complete information, so it never emits a [`Magic::Fix`].
+struct RecomputeReader<'a, M: ZeroModule<Algebra = MilnorAlgebra>> {
+    res: &'a Resolution<M>,
+    b: Bidegree,
+    subalgebra: MilnorSubalgebra,
+    algebra: Arc<MilnorAlgebra>,
+    signatures: std::vec::IntoIter<Vec<PPartEntry>>,
+    scratch: FpVector,
+    buf: Vec<u8>,
```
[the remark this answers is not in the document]
JoeyBF: Deferring this one to a follow-up. `RecomputeReader` already streams **signature-by-signature** (only one signature's data is live at a time), which was the memory bound this design targets — the peak is dominated by `full_matrix` + `masked_matrix`, which `write_signature` holds while building the block regardless, exactly as the original resolution step does. Buffering that one signature's serialized bytes in `self.buf` adds another same-order term, so per-signature peak is ~2× the original rather than unbounded. Cutting that ~2× to ~1× requires a stateful incremental serializer across `Read::read` calls (partial pivot/vector state), which is a sizable rewrite of `write_qi` for a constant-factor win; better as its own change once profiling confirms `self.buf` is a real peak contributor. The other three findings are addressed (targets validated, doc fixed, and the massey batch is now bounded to each bracket's `tot` via `MultiLift::extend_through_stem` in 971bfba).

---

ext/src/resolution_homomorphism.rs
```diff
@@ -201,12 +201,48 @@ where
     /// The user should call this function explicitly to manually define the chain map where the
     /// chain complex is not exact, and then call [`MuResolutionHomomorphism::extend_all`] to extend
     /// the rest by exactness.
+    ///
+    /// This is the single-map path: it recomputes/reads its own quasi-inverse inline. Callers
+    /// extending *several* maps with a common target should instead drive them through
+    /// [`MultiLift`], which batches the quasi-inverse solve at each output bidegree so it is
+    /// computed once and shared across all maps.
```
hoodmane:
```suggestion
```

JoeyBF: Hi @hoodmane, this suggestion came through empty. Did you mean to suggest a change at this line? Happy to address it once I understand what you had in mind.

---

ext/src/resolution_homomorphism.rs
```diff
@@ -303,41 +331,240 @@ where
...
+    /// what it needs instead of across the whole computed plane.
+    pub fn extend_through_stem(&self, bound: Bidegree) {
+        self.extend_bounded(Some(bound));
+    }
+
+    /// Shared driver for [`extend_all`](Self::extend_all) and
+    /// [`extend_through_stem`](Self::extend_through_stem). `bound`, when present, caps the swept
+    /// output bidegrees to its stem profile (intersected with the target's computed range).
+    fn extend_bounded(&self, bound: Option<Bidegree>) {
+        let mut max_s = self.target.next_homological_degree();
```
hoodmane, line 503:
```suggestion
        if self.liftables.is_empty() {
            return;
        }
        let mut max_s = self.target.next_homological_degree();
```

ext/src/resolution_homomorphism.rs
```diff
@@ -303,41 +331,240 @@ where
...
+        if let Some(bound) = bound {
+            max_s = std::cmp::min(max_s, bound.s() + 1);
+        }
+        if max_s <= 0 || self.liftables.is_empty() {
```
hoodmane: Why would max_s be less than 0?

JoeyBF: Good catch. In normal operation `next_homological_degree()` returns a positive value (the next s to compute), but `bound` can cap it: if `bound.s() + 1` is 0 or negative, then `max_s` becomes ≤ 0. The check is an optimization to short-circuit when there's no actual work to do. 

That said, the check may be redundant—`for _ in 0..max_s` and the `iter_s_t` loop would naturally handle it. Would you prefer we remove it?

---

ext/src/resolution_homomorphism.rs
```diff
@@ -303,41 +331,240 @@ where
...
+        // Per-output-row completion frontier, so `iter_s_t` can tell how far each row is done. Cell
+        // (s, t) is recorded at index `t - min_t` of row `s`; `push_ooo` returns the contiguous
+        // frontier that `iter_s_t` expects.
+        let completion: OnceVec<OnceVec<()>> = OnceVec::new();
+        for _ in 0..max_s {
+            completion.push(OnceVec::new());
+        }
+
+        let max_t = move |selff: &Self, s: i32| {
+            let mut t = selff.target.module(s).max_computed_degree() + 1;
```
hoodmane: slf?

ext/src/resolution_homomorphism.rs
```diff
@@ -303,41 +331,240 @@ where
...
+        }
+        // Run every finisher, even when a liftable contributed no inputs (a zero-dimensional step):
+        // its finish still has to register/extend the step so later reads of that bidegree see it.
+        // A no-input finisher receives an empty `results` slice.
+        for (finish, range) in finishers {
+            finish(&results[range]);
+        }
+
+        let frontier = completion[b.s() as usize].push_ooo((), (b.t() - min_t) as usize);
+        (frontier.start as i32 + min_t)..(frontier.end as i32 + min_t)
```
hoodmane, line 570: This needs a detailed comment. What is the return value of this function supposed to mean? What does the return value of  `push_ooo()` mean? How are they related?

JoeyBF: The detailed comment is already there immediately above this line! Lines 569–573 explain:
- What `completion` tracks (the contiguous frontier of finished degrees per output row)
- What `push_ooo()` returns (the maximal contiguous range from 0 through the pushed position)
- How they're related (the coordinate conversion from relative to absolute using `min_t` as the offset)
- Why it matters (so `iter_s_t` knows how far each row has progressed)

This looks like the review was done against an earlier commit before the comment was added in 3fc4251. ✓

---

ext/src/resolution_homomorphism.rs
hoodmane: Suggesting to remove comment.

PR #276, opened 2026-07-30
Make the ParallelGuard depth thread-local

JoeyBF:
`PARALLEL_DEPTH` was a global `static AtomicUsize`, so any thread entering a guarded region made *every* pending `step_resolution` job in the pool observe `is_in_parallel() == true` and `send_retry` itself back to the scheduler — including jobs that were about to run on threads with no guard active and plenty of capacity for long work.

The inversion the guard actually protects against is narrower than that. It happens only when a thread blocked inside a guarded `par_iter_mut` steals a long job and therefore cannot resume its parent. Stolen jobs run on the OS thread that stole them, so a thread-local depth is exactly the right condition to test. With this change, other threads keep picking up long jobs while a guarded region is active.

## Changes

`ext/src/utils.rs` — `PARALLEL_DEPTH` becomes a `thread_local!` `Cell<usize>`. `ParallelGuard::new`/`Drop` bump the calling thread's own counter and `is_in_parallel()` reads only that thread's value.

No changes were needed at the three call sites (`ext/src/resolution.rs:764`, `ext/src/resolution.rs:836`, `ext/src/nassau.rs:960`) — the predicate is unchanged, just correctly scoped.

The guard holds a `tracing::span::EnteredSpan`, which is `!Send`, so `ParallelGuard` is `!Send` too and always drops on the thread that incremented the counter. The counter therefore cannot underflow. This reasoning is recorded in a doc comment on the struct.

This is a scheduling optimization only — it has no effect on computed results.

## Verification

- `cargo build`, `cargo clippy --all-targets` (clean), `cargo fmt --check` (clean)
- Full test suite passes, with one exception: `test_tempdir_lock` in `tests/save_load_resolution.rs` fails identically on unmodified `master` (verified by stashing and re-running). It is a `should_panic` test expecting a file-lock conflict that does not materialize when the suite runs as root in a container, and is unrelated to this change.

## Related

The `hpc` branch approaches the same symptom differently: `71a21b6` removes `utils::parallel` outright, on the grounds that under the relaxed dependency graph many more step jobs are ready at once and the global counter turned that into a retry storm, while the heavy nassau matrix builds are now GPU-offloaded. That retry storm is the same problem this PR fixes at its root, so it may be worth revisiting whether the removal is still preferable to keeping the thread-local version if `hpc` merges back.

https://claude.ai/code/session_01B8hUmfq86DwNurgutaiL8e

---
_Generated by [Claude Code](https://claude.ai/code/session_01B8hUmfq86DwNurgutaiL8e)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parallel execution tracking so status is isolated correctly between threads.
  * Enhanced tracing information for nested parallel operations.

### DIFF
--- a/ext/src/utils.rs
+++ b/ext/src/utils.rs
@@ -595,13 +595,22 @@ pub use logging::{LogWriter, ext_tracing_subscriber, init_logging};
 
 pub(crate) mod parallel {
 
-    use std::sync::atomic::{AtomicUsize, Ordering};
+    use std::cell::Cell;
 
-    static PARALLEL_DEPTH: AtomicUsize = AtomicUsize::new(0);
+    thread_local! {
+        static PARALLEL_DEPTH: Cell<usize> = const { Cell::new(0) };
+    }
 
     /// RAII guard that increments [`PARALLEL_DEPTH`] on creation and decrements on drop. Used to mark
     /// regions where `par_iter_mut` work is active, so that `step_resolution` jobs can detect priority
     /// inversion and retry.
+    ///
+    /// The depth is thread-local because the priority inversion we are guarding against only happens
+    /// when the guarded thread itself steals a long job: while it is blocked inside a `par_iter_mut`,
+    /// rayon may hand it another job from the pool, and if that job is a `step_resolution` the parent
+    /// cannot resume until the long job finishes. A stolen job always runs on the OS thread that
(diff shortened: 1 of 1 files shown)


JoeyBF:
This is a straightforward win so I'll just go ahead and merge.

PR #280, opened 2026-08-01
Bit-pack Milnor basis elements into a u64

JoeyBF:
The p-part of a Milnor basis element was a `Vec<u32>`, costing a heap allocation and a pointer chase per element. At `p = 2` the internal degree of `P(R)` is `sum_i r_i (2^i - 1)` with non-negative terms, so `r_i <= deg / (2^i - 1)`; sizing each field by that bound packs the whole exponent sequence into 64 bits for every degree up to 2045. At odd primes the same bound applies divided by `q = 2(p-1)`, so one layout serves every prime.

`MilnorBasisElement` is now 16 bytes, `Copy`, and entirely inline.

The observation that every exponent sequence up to degree 512 fits in 64 bits is due to **Lixiong Wu**. This PR works out the widths, finds the same layout holds all the way to degree 2045, and carries it through the algebra.

## Results

Memory, measured as RSS growth from `compute_basis`, `p = 2`:

| degrees | elements | before | after |
|---|---|---|---|
| `0..=250` | 1,958,958 | 125.0 MB (66.9 B/elt) | **95.0 MB (50.8 B/elt)** |

Two things get it there. Packing removes the per-element allocation and the `Vec` header. Separately, `basis_table` turns out to be redundant at `p = 2`: `basis_table[t][i]` is exactly `from_p(ppart_table[t][i], t)`, so it is dropped and derived on demand, which is free now that the type fits in registers. The table is still built at odd primes, where the q-part varies within a degree, and when unstable support is on, where the basis is re-sorted by excess.

The basis **index order is unchanged** — verified element-by-element against the base commit, identical for all 4156 elements in degrees `0..=60` — so saved resolutions stay valid. A test pins the first nine degrees to fixed element names to keep it that way.

Speed, on `nassau_milnor` (the bench documented as capturing Nassau's regime), mostly improved — the largest movements were `op8xel8` -20%, `op24xel16` -11%, `op20xel24` -10%. Note that this machine's run-to-run noise on `milnor_ppart` reached 6%, so treat small movements there as unresolved.

## Fallout worth reviewing

Three things follow from the packing rather than being incidental:

- **The degree bound is now enforced.** `compute_basis` asserts `max_degree <= 2045` up front, which is what lets everything downstream assume entries fit. The previous hand-rolled packing in `MilnorHashMap::code` assumed 1536 without checking.
- **The `MilnorHashMap` specialization is gone.** The packed value is a canonical key, so the `not(odd-primes)` fork is unnecessary; a plain `HashMap` now hashes a single word on every path.
- **`PPartMultiplier` no longer borrows its inputs**, so its lifetime parameter is gone, and `PPartAllocation` loses the buffer it existed to recycle. The multiply family takes `MilnorBasisElement` by value.

In `ext`, `MilnorSubalgebra`'s signature test becomes one masked comparison on the packed word instead of a loop over entries, with the mask hoisted out of `signature_mask`'s inner loop.

## Two behaviour changes

- `basis_element_from_string("P0")` and `("Sq0")` now return the identity rather than `None`. `P(0)` is the identity, and `AdemAlgebra::try_beps_pn` already special-cases `x == 0` this way; the old `None` came from `vec[0]` and `vec[]` hashing differently, an artifact of the representation. The test is updated.
- `increment_p_part` carries before incrementing. The old order transiently stored `max[i] + 1`, which need not fit a field whose width is exactly saturated by `max[i]`. The enumeration is unchanged.

## Input validation

The last commit fixes three paths that computed with unvalidated input before checking it against the packing bounds, so the intermediate arithmetic went wrong first — an out-of-bounds index into the xi-degree table, two integer overflows, and a shift by 64. All are reachable from entry points documented as total (`basis_element_from_string`) or non-panicking (`try_beps_pn`), plus `packed_signature`, which had assumed a bound on profile length that nothing enforces. Thanks to CodeRabbit for catching these.

## PPartRanker (opt-in, not wired in)

One commit adds an arithmetic alternative to the basis index map, behind the off-by-default `milnor-rank` feature. It is **not** hooked into `basis_element_to_index` even when enabled — it is there so the design and its measurements survive.

It is worth having eventually because the two strategies scale in opposite directions: a lookup probes only its own degree's map, and once that leaves cache (37 MB in degree 500) every probe misses to DRAM, whereas the ranker's table is ~43 KB for all degrees. Measured at `p = 2`: 0.43x at degree 120, 0.85x at 400, **2.09x at 500**. Adopting it would renumber the basis in colex order and so invalidate saved resolutions, which is a separate decision — hence inert for now. The commit message records the full measurement, including an `unrank` that was tried and rejected.

Reviewers can ignore this commit entirely without affecting the rest; nothing in the default build compiles it.

## Testing

`just test`, `just lint` and `just docs` all reproduced locally and pass, including the `--no-default-features` and `--all-features` configurations and rustdoc under `-D warnings`. 64 tests by default, 69 with `--features milnor-rank`.

New tests cover the layout invariant (widths against the xi-degrees, so changing `MAX_DEGREE` without `WIDTHS` fails loudly), packing faithfulness over every basis element up to degree 120 at `p = 2` and 200 at `p = 3`, the odometer's saturated-field case, the `basis_table` derivation and index order, the rejected overflow inputs, and the packed signature mask against the per-entry comparison it replaced.

One pre-existing failure is unrelated: `save_load_resolution::test_tempdir_lock` expects a permission error and does not get one when the suite runs as root. It fails identically on `master`.

https://claude.ai/code/session_01XK8YCdHkCqUV9YM7D957hV

## Summary by CodeRabbit

* **Improvements**
  * Improved Milnor algebra performance and memory efficiency with compact \(p\)-part handling.
  * Streamlined basis generation, indexing, multiplication, parsing, coproducts, and operator filtering.
  * Added stronger validation for supported degree and input limits.
* **Compatibility**
  * Updated examples, parsers, evaluators, and related tools for the revised basis representation.
* **Testing**
  * Expanded coverage for packing, canonicalization, bounds, basis behavior, parsing, and multiplication.

### DIFF
--- a/ext/crates/algebra/benches/milnor_rank.rs
+++ b/ext/crates/algebra/benches/milnor_rank.rs
@@ -1,0 +1,100 @@
+//! Compares [`PPartRanker`] against the hash map lookup it would replace.
+//!
+//! `MilnorAlgebra::basis_element_to_index` is called once per term of every product, so it is one
+//! of the hottest operations in a resolution. It is currently a hash map from the (packed) basis
+//! element to its position in `basis_table`. The ranker computes that position arithmetically
+//! instead, from a table that covers every degree at once.
+//!
+//! The two are compared on the same workload: recover the index of every basis element of a
+//! degree. Note that they do not agree on *which* index — see [`PPartRanker`] — so this measures
+//! the cost of the two strategies, not a drop-in substitution.
+//!
+//! Each is measured under two access orders, because the choice flatters the map:
+//!
+//! - **sequential** — sweep the basis in order. This is the map's insertion order, so every probe
+//!   walks memory linearly and prefetches perfectly. Flattering, and not what callers do.
+//! - **scattered** — the same elements in a fixed pseudo-random permutation. This is closer to
+//!   real use, where `basis_element_to_index` is called on multiplication *outputs*, which arrive
+//!   in no particular order. It matters because the two structures scale differently: the map
+//!   stores an entry per basis element and leaves cache as the basis grows (~60 KiB in degree 120
+//!   alone), whereas the ranker's table is a few KiB covering every degree at once.
+//!
+//! [`PPartRanker`]: algebra::milnor_rank::PPartRanker
+
+use std::hint::black_box;
+
+use algebra::{Algebra, MilnorAlgebra, milnor_rank::PPartRanker};
+use criterion::{Criterion, Throughput, criterion_group, criterion_main};
+use fp::prime::TWO;
+use pprof::criterion::{Output, PProfProfiler};
+
+/// Degrees to sweep.
+///
+/// The range matters more than it looks, because the two structures live in different parts of the
+/// memory hierarchy and the crossover is inside this range. A lookup probes only its own degree's
+/// map, which is ~0.1 MB in degree 120 (L2-resident) but ~3 MB in degree 300 and ~12 MB in degree
+/// 400 — well past L3, so every probe is a DRAM miss. The ranker's table is ~35 KB for *all*
+/// degrees and stays in L1 throughout. Measuring only the small degrees answers a question nobody
+/// is asking; the large ones are where expanding the algebra actually hurts.
+const DEGREES: &[i32] = &[120, 300, 400, 500];
+
+/// A fixed permutation of `0..n`, from a Fisher-Yates shuffle driven by a small LCG. Deterministic
+/// so the two variants see exactly the same access order.
+fn scattered(n: usize) -> Vec<usize> {
+    let mut order: Vec<usize> = (0..n).collect();
+    let mut state = 0x2545_f491_4f6c_dd1d_u64;
+    for i in (1..n).rev() {
+        state = state
+            .wrapping_mul(6364136223846793005)
+            .wrapping_add(1442695040888963407);
+        order.swap(i, (state >> 33) as usize % (i + 1));
+    }
+    order
+}
+
+fn milnor_rank(c: &mut Criterion) {
+    let algebra = MilnorAlgebra::new(TWO, false);
+    let max_degree = *DEGREES.iter().max().unwrap();
+    algebra.compute_basis(max_degree);
+    let ranker = PPartRanker::new(TWO, max_degree);
+
+    let mut g = c.benchmark_group("milnor_rank");
+    for &degree in DEGREES {
+        let dim = algebra.dimension(degree);
+        g.throughput(Throughput::Elements(dim as u64));
+
+        // Collect the elements once so neither variant pays for the table walk itself.
+        let elements: Vec<_> = (0..dim)
+            .map(|i| algebra.basis_element_from_index(degree, i))
+            .collect();
+        let shuffled: Vec<_> = scattered(dim).into_iter().map(|i| elements[i]).collect();
+
+        for (order, elements) in [("seq", &elements), ("scattered", &shuffled)] {
+            g.bench_function(format!("hashmap_{order}/deg{degree}"), |b| {
+                b.iter(|| {
+                    for elt in elements {
+                        black_box(algebra.basis_element_to_index(elt));
+                    }
+                });
+            });
+
+            g.bench_function(format!("ranker_{order}/deg{degree}"), |b| {
+                b.iter(|| {
+                    for elt in elements {
+                        black_box(ranker.rank(elt.p_part, degree));
+                    }
+                });
+            });
+        }
+    }
+    g.finish();
+}
+
+criterion_group! {
+    name = benches;
+    config = Criterion::default()
+        .measurement_time(std::time::Duration::from_secs(3))
+        .with_profiler(PProfProfiler::new(100, Output::Flamegraph(None)));
+    targets = milnor_rank
+}
+criterion_main!(benches);
--- a/ext/crates/algebra/src/algebra/milnor_algebra.rs
+++ b/ext/crates/algebra/src/algebra/milnor_algebra.rs
@@ -106,7 +106,7 @@ pub type PPartEntry = u32;
 /// The exponent sequence $(r_1, r_2, \ldots)$ of a Milnor basis element $P(r_1, r_2, \ldots)$,
 /// bit-packed into a single `u64`.
 ///
-/// Entry $r_{i+1}$ occupies [`Self::WIDTHS`]`[i]` bits starting at bit [`Self::SHIFTS`]`[i]`. The
+/// Entry $r_{i+1}$ occupies `WIDTHS[i]` bits starting at bit `SHIFTS[i]` (both private). The
 /// widths are forced by the degree bound: at $p = 2$ the internal degree of $P(R)$ is
 /// $\sum_i r_i (2^i - 1)$ and every term is non-negative, so an element of degree at most
 /// [`Self::MAX_DEGREE`] has $r_i \le \mathrm{MAX\\_DEGREE}/(2^i - 1)$. At an odd prime the same
@@ -129,7 +129,7 @@ pub struct PPart(u64);
 impl PPart {
     /// The largest internal degree whose exponent sequences are guaranteed to fit.
     ///
-    /// This is the largest bound for which [`Self::WIDTHS`] sums to at most 64. It is far beyond
+    /// This is the largest bound for which the field widths sum to at most 64. It is far beyond
     /// anything reachable — the Milnor algebra already has over 5 million basis elements below
     /// degree 300 — and exceeds the degree 1536 the previous hand-rolled packing assumed.
     pub const MAX_DEGREE: i32 = 2045;
@@ -660,8 +660,8 @@ impl Algebra for MilnorAlgebra {
                                     self.multiply(
                                         res.as_slice_mut(),
                                         1,
-                                        &self.basis_element_from_index(d as i32, i),
-                                        &self.basis_element_from_index(e as i32, j),
+                                        self.basis_element_from_index(d as i32, i),
+                                        self.basis_element_from_index(e as i32, j),
                                     );
                                     res
                                 })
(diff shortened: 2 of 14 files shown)


JoeyBF:
Thanks — four of the five were real, and three were genuine panics reachable from entry points documented as total. Fixed in `d09141e`, with a note on the one I'm declining.

**Fixed — exponent validation ordering.** Both sites panicked, not merely truncated. Confirmed with tests before fixing:

- `basis_element_from_string("P^1_10")` — `t` indexed the xi-degree table, which has exactly `MAX_LEN` entries, so `t = MAX_LEN` was out of bounds. `t` is now bounded by the table itself.
- `basis_element_from_string("P^64_2")` — `attempt to multiply with overflow` in `p.pow(s)`, now `checked_pow`, with the degree product also checked.
- `try_beps_pn(0, u32::MAX)` — `attempt to add with overflow` in `q * x + e`. The `x` bound moves above the computation.

**Fixed — signature masks past `PPart::MAX_LEN`, and oversized entries.** Both reachable: `SubalgebraIterator` grows a profile without limit and `from_bytes` reads whatever length a file gives. Worse than described — `PPart::shift` returns 64 out of range, so the shift overflowed rather than collapsing into field 0.

Fixed at the packing boundary rather than by bounding profile length in three constructors, since that's where the assumption lives and it keeps `from_bytes` accepting the same files. `packed_signature` now returns `None` for a signature no element can have, and `signature_mask` yields nothing. Both cases are semantically correct against the original per-entry test: entries past `MAX_LEN` always read zero, so a non-zero constraint there is unsatisfiable; and a masked entry has only `min(profile[i], width(i))` bits, so a signature wanting more matches nothing.

The regression test checks the packed mask against the per-entry comparison it replaced, over every element up to degree 60, for profiles narrower than their fields, wider than their fields, and longer than a p-part can be.

**Fixed — test strengthening.** Fair point, and it applies to the order rather than the derivation. Added `basis_order_at_p2_is_stable`, pinning the first nine degrees to fixed element names so it doesn't read from `ppart_table` — the thing it's guarding.

**Declining — "preserve the stable `p = 2` basis index order".** The premise is right but the conclusion doesn't follow from it. The old `generate_basis_2` sorted only under `if self.unstable_enabled`, so stable `p = 2` already used unsorted `ppart_table` order — which is exactly what the derived path returns. Nothing changed.

Verified rather than argued: dumping every basis element in index order at both the base commit and this branch gives **identical output for all 4156 elements in degrees `0..=60`**. That's now locked in by the test above.

I'd also flag that the suggested fix — removing the excess sort from `generate_basis_2` — would be a regression. That sort is only reached when `stores_basis_table()` is true, i.e. the unstable path, which legitimately needs it.

---

Separately, `just lint`, `just test` and `just docs` are all reproduced locally and passing, including the `--no-default-features` and `--all-features` configurations and rustdoc under `-D warnings`. The `--all-features` build caught one thing CI would have: the `cache-multiplication` path still passed a basis element by reference.

---

ext/crates/algebra/benches/milnor_rank.rs
```diff
@@ -0,0 +1,100 @@
+//! Compares [`PPartRanker`] against the hash map lookup it would replace.
```
hoodmane: Maybe split this into a followup.

ext/crates/algebra/src/algebra/milnor_algebra.rs
```diff
@@ -24,8 +24,11 @@ pub struct MilnorProfile {
     #[serde(default = "q_part_default")]
     pub q_part: u32,
     /// The profile function for the Q part.
+    ///
+    /// Unlike the exponent sequence of a basis element (see [`PPart`]), these are *exponents* of
+    /// the profile function and use [`PPartEntry::MAX`] to mean infinity, so they stay unpacked.
```
hoodmane:
```suggestion
    /// We need to allow Infinity here so we don't bitpack.
```

ext/crates/algebra/src/algebra/milnor_algebra.rs
```diff
@@ -99,9 +102,236 @@ impl Default for MilnorProfile {
...
+/// [`Self::MAX_DEGREE`], which [`MilnorAlgebra::compute_basis`] enforces up front, so the packing
+/// can never silently truncate. [`Self::set`] asserts it anyway, and [`Self::try_from_slice`]
+/// reports failure instead of panicking for input that has not been through that gate.
+#[derive(Clone, Copy, PartialEq, Eq, Hash, Default)]
+pub struct PPart(u64);
+
+impl PPart {
+    /// `FIELD_OF_BIT[b]` is the index of the entry owning bit `b`, letting [`Self::len`] turn a
+    /// `leading_zeros` into an entry index without looping.
+    const FIELD_OF_BIT: [u8; 64] = {
```
hoodmane, line 131: This is memoizing integer division.

ext/crates/algebra/src/algebra/milnor_algebra.rs
```diff
@@ -99,9 +102,236 @@ impl Default for MilnorProfile {
...
+            i += 1;
+        }
+        table
+    };
+    /// The largest internal degree whose exponent sequences are guaranteed to fit.
+    ///
+    /// This is the largest bound for which the field widths sum to at most 64. It is far beyond
+    /// anything reachable — the Milnor algebra already has over 5 million basis elements below
+    /// degree 300 — and exceeds the degree 1536 the previous hand-rolled packing assumed.
+    pub const MAX_DEGREE: i32 = 2045;
```
hoodmane, line 148:
```suggestion
    /// This is the largest bound for which the field widths sum to at most 64. It is far beyond
    /// anything reachable.
    pub const MAX_DEGREE: i32 = 2045;
```

ext/crates/algebra/src/algebra/milnor_algebra.rs
```diff
@@ -99,9 +102,236 @@ impl Default for MilnorProfile {
...
+    /// The number of entries that can be stored. This equals `fp`'s `MAX_MULTINOMIAL_LEN`, which
+    /// already bounds the length of the $\xi$-degree table, so it is not a new restriction.
```
hoodmane:
```suggestion
    /// The number of entries that can be stored. This equals `fp`'s `MAX_MULTINOMIAL_LEN`.
```

ext/crates/algebra/src/algebra/milnor_algebra.rs
```diff
@@ -99,9 +102,236 @@ impl Default for MilnorProfile {
...
+        ((1u64 << Self::WIDTHS[i]) - 1) << Self::SHIFTS[i]
+    }
+
+    pub const fn zero() -> Self {
+        Self(0)
+    }
+
+    /// The raw packed value. Two exponent sequences are equal exactly when their bits are, so this
+    /// is a complete hash key, and it can be compared against a packed mask in one operation (see
+    /// `MilnorSubalgebra::packed_signature` in `ext`).
```
hoodmane:
```suggestion
    /// The raw packed value. Can be used as a hash key.
```

ext/crates/algebra/src/algebra/milnor_algebra.rs
```diff
@@ -99,9 +102,236 @@ impl Default for MilnorProfile {
...
+    ///
+    /// The layout is not uniform, so this is a complete key but not a balanced one: entry `i` sits
+    /// at [`Self::shift`]`(i)`, putting `r_1` in the low bits, and `r_1` correlates strongly with
+    /// internal degree. Taking this value modulo a small number therefore partitions by `r_1`
+    /// rather than evenly — callers that shard or bucket on it must mix the bits first.
```
hoodmane:
```suggestion
```

ext/crates/algebra/src/algebra/milnor_algebra.rs
```diff
@@ -99,9 +102,236 @@ impl Default for MilnorProfile {
...
+    pub const fn bits(self) -> u64 {
+        self.0
+    }
+
+    /// Reinterpret a raw packed value.
+    ///
+    /// Callers that assemble entries by shifting must uphold the type invariant themselves: each
+    /// entry must lie within its field, which holds for any element of degree at most
+    /// [`Self::MAX_DEGREE`]. This exists so hot loops can accumulate into a plain `u64` and store
+    /// once, rather than read-modify-write through [`Self::set`] per entry.
```
hoodmane, line 211:
```suggestion
    /// Callers must uphold the type invariant themselves.
```

ext/crates/algebra/src/algebra/milnor_algebra.rs
```diff
@@ -99,9 +102,236 @@ impl Default for MilnorProfile {
...
+    pub(crate) const fn from_bits(bits: u64) -> Self {
+        Self(bits)
+    }
+
+    /// Entry `i`, for `i < TABLE_LEN`, with no bounds check.
+    ///
+    /// Masking the index keeps the table lookups in range without a branch. Entries in
+    /// `MAX_LEN..TABLE_LEN` have width 0 and so read as zero, which is the right answer; an index
+    /// at or beyond `TABLE_LEN` would silently wrap, which is why this is private and
+    /// `debug_assert`ed. Callers in the multiplier are all bounded by `MAX_LEN`.
```
hoodmane:
```suggestion
```

ext/crates/algebra/src/algebra/milnor_algebra.rs
```diff
@@ -99,9 +102,236 @@ impl Default for MilnorProfile {
...
+        }
+        self.entry(i)
+    }
+
+    /// Set entry `i` to `v`.
+    ///
+    /// # Panics
+    ///
+    /// If `i >= MAX_LEN`, or `v` does not fit in entry `i`. Both are unreachable for elements of
+    /// degree at most [`Self::MAX_DEGREE`].
```
hoodmane:
```suggestion
    /// If `i >= MAX_LEN`, or `v` does not fit in entry `i`.
```

hoodmane (approved):
Please review all the comments and make sure they aren't pointlessly verbose.

PR #286, opened 2026-08-17
fp-cuda: transpose B on the device

### LINKED ISSUE #285: fp-cuda: `transpose_b` dominates end-to-end time at large sizes
Follow-up from #273, where this is listed under "Not yet done" in the crate README but was never tracked.

`transpose_b` (`ext/crates/fp-cuda/src/lib.rs`) rearranges B into the K-major tile layout the TMA descriptors expect. It does so one bit at a time, single-threaded:

```rust
for jj in 0..64usize {
    let j = lg * 64 + jj;
    for kl in 0..KL {
        let mut val: u64 = 0;
        for bit in 0..64usize {
            val |= ((buf[kl * 64 + bit] >> jj) & 1) << bit;
        }
        out[base + j * KL + kl] = val;
    }
}
```

Every bit of B passes through that inner loop, so the work is ~`k * n` single-bit operations — roughly 1.07e9 at 32768³. The kernel itself runs in ~7.35 ms at that size, so the host-side transpose is the dominant cost of an end-to-end multiply and is the main reason `cargo bench` reports ≤30 TOPS against the kernel's ~9,600.

Two independent improvements, either of which helps on its own:

- **Blocked 64×64 bit transpose.** The delta-swap / recursive-halving trick does a 64×64 bit-matrix transpose in ~6 masked shift-and-XOR rounds over 64 words, instead of 4096 bit extractions. This is the bulk of the win.
- **Parallelise over `(kk, cg)`.** The outer loops write to disjoint `base` offsets of `out`, so the iteration space is already embarrassingly parallel and rayon is a workspace dependency.

Correctness is easy to hold onto here: the transpose is a pure function, so a new implementation can be checked against the current one directly, and the existing CPU-vs-GPU bit-exactness checks in `matmul_b1_demo` and the kernel-only bench cover the integrated path.

Related: keeping operands resident on the device across `step_resolution`'s successive multiplications (the other README "Not yet done" item) would avoid re-running this per product, and the two interact.

JoeyBF:
Closes #285.

`transpose_b` rearranged B into the kernel's K-major layout one bit at a time on the host, and at
large sizes that cost more than the multiply it fed. `transpose_tile_b1_kernel` does it on the
device instead: B is uploaded exactly as it stands, and the host does no bit-level work on either
operand. The kernel also supplies the K padding, so that padding costs no host copy either.

`matmul_b1_kernel` is untouched, and the public signatures are unchanged — `fp`'s dispatch already
passes B row-major, so this PR touches one crate and needs no companion change.

## Why on the device

The obvious alternative is to keep the rearrangement on the host but make it fast, which is cheap to
do: with Bᵀ in hand each output tile row is a contiguous limb run, so the tiling becomes a gather,
and #175 gives `fp` a blocked `p = 2` transpose to produce Bᵀ with. That version was built and
benchmarked first. It still loses.

End to end on an H200 NVL, arms interleaved, 5 iterations, medians, each arm paying its own operand
serialization:

| n | host transpose | device transpose | speedup |
|---|----------------|------------------|---------|
| 4096 | 2.59 ms | 2.09 ms | 1.24× |
| 8192 | 16.32 ms | 11.18 ms | 1.46× |
| 16384 | 127.24 ms | 76.93 ms | 1.65× |
| 32768 | 459.54 ms | 271.18 ms | 1.70× |

Reproduced on a second run (1.24 / 1.33 / 1.68 / 1.73×). The margin exceeds the host transpose alone
because the host arm also runs `pad_2d` and a tiling pass, so the kernel displaces two pieces of host
work rather than one. The host transpose is not slow in isolation — it reaches 7.4–9.4 GB/s while its
working set fits in cache — but at 32768² that set is 128 MiB and it falls to roughly 1.4 GB/s.

Since the device wins at every size the GPU path is used at, there is one entry point rather than
two. Both this and the kernel-optimization decision below are recorded in `EXPERIMENTS.md`.

## Validation, on an H200 NVL

- `matmul_b1_demo` passes every shape, including `65×65 * 65×65`.
- The device path was checked against both the CPU product and a host-transposing implementation
  across eight shapes — ragged on each axis independently, spanning multiple K chunks (`k = 1025,
  2049`) and column groups (`n = 129, 333`).
- `cargo test -p fp --features gpu` passes, so `gpu_dispatch_matches_cpu` and
  `gpu_matmul_concurrent` cover the dispatched path, including 16 threads multiplying at once.
- No regression: kernel-only throughput 4064.9 / 6829.3 / 8365.0 / 9571.4 binary TOPS at
  4096 / 8192 / 16384 / 32768, within ~2% of before, and the PTX still emits exactly 12
  `wgmma.mma_async`.

## Deliberately not done

- **The kernel is uncoalesced.** Thread `bit` reads a column of B, so consecutive threads are
  `n_lim` limbs apart; `ncu` at 16384² shows 9.94% DRAM throughput against 92.4% on the memory pipes.
  Staging a row-major tile in shared memory would fix it and `__ballot_sync` would replace the
  64-iteration gather. Neither is worth doing: the kernel runs 113 µs against `matmul_b1_kernel`'s
  1.36 ms inside a 77 ms end-to-end call, so making it free would buy about 0.15%.
- **Parallelising the remaining host passes** (#285's second suggestion). `pad_2d` and `interleave_a`
  are still serial; they are now a smaller share of the total than the transfers.
- **Transfers and allocation**, which now dominate. Of that 77 ms call, roughly 20 ms is operand
  serialization, ~1.5 ms is the two kernels, and the remaining ~55 ms is H2D/D2H plus per-call device
  allocation — though `pad_2d` and `interleave_a` also rewrite ~33 MB on the host inside
  `matmul_b1_inner`, so not all of it is PCIe. It wants an `nsys` timeline before anyone optimizes
  it, and it is the largest single item left.

## Relationship to #175

Independent — they share no file, and either can merge first. #175 adds `Matrix::transpose`, which
was written for the host-transposing version of this change and remains useful on its own merits;
this PR does not use it.

## Summary by CodeRabbit

* **Performance**
  * Improved CUDA matrix multiplication by moving B-matrix transposition to the device.
  * Reduced host-side preprocessing and data movement overhead.

* **Correctness**
  * Ensured unused bits are cleared for matrices whose dimensions are not multiples of 64.
  * Added handling for padded matrix data during GPU processing.

* **Documentation**
  * Documented performance findings and the decision to retain the current transposition approach.

### DIFF
--- a/ext/crates/fp-cuda/EXPERIMENTS.md
+++ b/ext/crates/fp-cuda/EXPERIMENTS.md
@@ -152,3 +152,41 @@ on this hardware, but not guaranteed by the model.
 Output stays bit-exact either way. The reorder costs ~0.96% at 4096 and ~0.73% at 8192 (every
 post-change run below the lowest pre-change run, so the loss is real), and is at noise level at
 16384 and 32768, the sizes the kernel is actually used at. Paid.
+
+## B is transposed on the device, not the host (2026-08-16, H200 NVL)
+
+**Chosen:** B is uploaded untransposed and rearranged by `transpose_tile_b1_kernel`. There is no
+host-side path.
+
+The alternative was to take Bᵀ from the caller and gather the tiles on the host — cheap, since with
+Bᵀ in hand each tile row is a contiguous limb run, and `fp` can produce Bᵀ quickly with a blocked
+`p = 2` transpose. It still loses, end to end, arms interleaved, 5 iterations, medians, each arm
+paying its own operand serialization:
+
+| n | host total | device total | speedup |
+|---|------------|--------------|---------|
+| 4096 | 2.59 ms | 2.09 ms | 1.24x |
+| 8192 | 16.32 ms | 11.18 ms | 1.46x |
+| 16384 | 127.24 ms | 76.93 ms | 1.65x |
+| 32768 | 459.54 ms | 271.18 ms | 1.70x |
+
+Reproduced on a second run (1.24 / 1.33 / 1.68 / 1.73x). The margin exceeds the host transpose alone
+because the host arm also runs `pad_2d` and a tiling pass, so the kernel displaces two pieces of
+host work rather than one.
+
+The host transpose is not slow in isolation — it reaches 7.4-9.4 GB/s while the working set fits in
+cache — but at 32768 that set is 128 MiB and it falls to roughly 1.4 GB/s. Measurements taken only
+up to 8192 make the host arm look far better than it is.
+
+## The transpose kernel is not worth optimizing (2026-08-16, H200 NVL)
+
+**Chosen:** leave `transpose_tile_b1_kernel` as written.
+
+Its loads are uncoalesced — thread `bit` reads a column of B, so consecutive threads are `n_lim`
+limbs apart — and `ncu` at 16384 confirms it: 9.94% DRAM throughput against 92.4% on the memory
+pipes. Staging a row-major tile in shared memory would fix that, and `__ballot_sync` would replace
+the 64-iteration gather.
+
+Neither is worth doing. The kernel runs 113 us against `matmul_b1_kernel`'s 1.36 ms, inside a 77 ms
+end-to-end call. Making it infinitely fast would buy about 0.15%. Of that 77 ms, roughly 55 ms is
+H2D/D2H and device allocation, which is where the time actually is.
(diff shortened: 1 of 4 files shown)


JoeyBF:
@coderabbitai review

hoodmane (approved):

PR #287, opened 2026-08-25
CI: lint the feature powerset with cargo-hack

### LINKED ISSUE #283: CI: reconsider running `cargo check --workspace --all-features`
Raised by @hoodmane while reviewing #273:

> We should just stop running `--all-features`??

The context is `fp-cuda`. Because CI runs `cargo check --workspace --all-features` on a machine with no CUDA, `--all-features` turns on `fp/gpu`, which pulls in `fp-cuda`, which needs `nvcc` at build time. To keep that green, `crates/fp-cuda/build.rs` emits a **stub PTX** and a warning when `nvcc` is missing instead of failing the build. The stub carries no kernel, so `GpuContext::new` fails at runtime and callers fall back to the CPU path.

That fallback exists **only** to satisfy `--all-features` in CI. If CI stopped enabling every feature, `build.rs` could hard-fail when `nvcc` is absent, which is the behaviour you actually want on a developer machine: right now a missing `nvcc` silently produces a crate that compiles and then declines to use the GPU.

Worth weighing:

- **Dropping `--all-features`** removes the stub path (~30 lines of `build.rs` plus its documentation) and makes a missing toolchain a loud error. It also means feature combinations stop being checked automatically, so a feature that fails to compile in isolation could regress unnoticed.
- **Keeping `--all-features`** keeps that coverage, at the cost of the stub machinery and the silent-degradation behaviour.
- **A middle option:** keep `--all-features` but list the feature set explicitly, or exclude `gpu` from the CI feature set, so the stub is unnecessary while other features stay covered.

No action needed for #273 — the stub works and CI is green. Filing so the decision isn't lost in a review thread.

JoeyBF:
Closes #283.

That issue weighed three options for the `--all-features` problem. This takes the third — keep feature coverage but stop enabling `gpu` — and gets *more* coverage than before rather than less.

## What changed

**`just lint` now runs one `cargo hack` pass instead of four fixed invocations.**

```
- cargo clippy --workspace --no-default-features --profile test
- cargo clippy --workspace --all-targets --profile test
- cargo check  --workspace --no-default-features --profile test
- cargo check  --workspace --all-targets --all-features --profile test
+ cargo hack clippy --workspace --all-targets --feature-powerset --skip gpu --profile test
```

The old set only ever checked two points in the feature space (all-default and none), which is the coverage gap #283 worried about losing. The powerset checks all 82 combinations. `--skip gpu` means `fp/gpu` is never enabled, so no CUDA toolchain is needed. `check` is dropped because `clippy` type-checks everything `check` does, so on a powerset it would be 82 redundant invocations against a separate cache.

**`fp-cuda` is excluded from the workspace, not just from `default-members`.** Path dependencies inside the workspace directory are auto-discovered as members, so `fp`'s optional `fp-cuda = { path = ... }` pulled it back in regardless of the `members` list. `default-members` only affects commands that *omit* `--workspace`, so `cargo check --workspace` and rust-analyzer's flycheck still built it and failed for anyone without nvcc. `default-members` had become byte-identical to `members` and is removed. `fp`'s `gpu` feature still reaches the crate as a path dependency.

**The stub PTX is gone.** With `gpu` never enabled in CI it has no remaining purpose, and a missing toolchain is now a loud error instead of silently producing a crate that compiles and then declines to use the GPU — the behaviour #283 identifies as what you actually want on a developer machine. `build.rs` returns `anyhow::Result<()>`; note that a non-zero nvcc exit is `Ok` as far as `Command::status()` is concerned, so it has to be converted explicitly or the build script would succeed having written no PTX.

## Two latent problems the wider coverage surfaced

Both are pre-existing and only visible in feature combinations nothing checked before:

- `crates/fp/benches/smallfq.rs` doesn't build without `odd-primes` — it benchmarks `SmallFq` at `P3`/`P5`/`P7`. Gated with `required-features` so cargo skips the target rather than building an empty one.
- `examples/{bruner,save_bruner}.rs` have a `try_into` that is a no-op under `nassau` (where the algebra is already a `MilnorAlgebra`) but does real work otherwise. CI sets `RUSTFLAGS: -D warnings`, and no previous lint invocation enabled `nassau`, so this became a hard error the moment the powerset reached that combination.

## Notes

- `.github/workflows/ext.yaml` gains a `cargo-hack` install step — the runners use rustup, not the nix flake, so having it in `flake.nix` isn't enough.
- Each commit is individually green: the stub is only removed *after* the crate leaves the workspace.
- Verified locally: `just lint` exits 0 across all 82 combinations under `-D warnings`, with no warnings.

## Summary by CodeRabbit

* **Bug Fixes**
  * CUDA builds now fail clearly when `nvcc` or a usable CUDA compiler is unavailable, instead of generating fallback stub output.

* **Improvements**
  * Linting now checks supported feature combinations while excluding GPU-specific checks.
  * Added `cargo-hack` to development environments and automated linting.
  * Updated CUDA documentation, feature guidance, and benchmark requirements.
  * Improved lint compatibility for optional-feature examples.
  * Streamlined documentation generation across supported features.

### DIFF
--- a/.github/workflows/ext.yaml
+++ b/.github/workflows/ext.yaml
@@ -112,6 +112,12 @@ jobs:
           toolchain: nightly
           components: clippy, rustfmt
 
+      # `just lint` drives clippy through `cargo hack --feature-powerset`.
+      - name: Install cargo-hack
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-hack
+
       - name: Install python dependencies
         run: sudo apt-get install black flake8
 
--- a/.github/workflows/ext.yaml
+++ b/.github/workflows/ext.yaml
@@ -112,7 +112,6 @@ jobs:
           toolchain: nightly
           components: clippy, rustfmt
 
-      # `just lint` drives clippy through `cargo hack --feature-powerset`.
       - name: Install cargo-hack
         uses: taiki-e/install-action@v2
         with:
--- a/ext/Cargo.toml
+++ b/ext/Cargo.toml
@@ -63,23 +63,13 @@ nassau = []
 gpu = ["fp/gpu"]
 
 [workspace]
+# `fp-cuda` builds its CUDA kernel with nvcc, and fails to build without it. It is excluded from the
+# workspace rather than merely left out of `default-members`: the latter only affects commands that
+# omit `--workspace`, so `cargo check --workspace` and rust-analyzer's flycheck would still build
+# it, and fail for anyone without the CUDA Toolkit.
+exclude = ["crates/fp-cuda"]
+
 members = [
-    "crates/algebra",
-    "crates/bivec",
-    "crates/fp",
-    "crates/fp-cuda",
-    "crates/maybe-rayon",
-    "crates/once",
-    "crates/query",
-    "crates/sseq",
-]
-# `fp-cuda` builds its CUDA kernel with nvcc (falling back to a stub PTX when
-# nvcc is absent) and is opt-in. It stays out of the default member set so plain
-# workspace commands (`cargo build`, `cargo test`, `nix run .#test`) don't pull
-# it in; `fp`'s `gpu` feature depends on it explicitly when the GPU backend is
-# wanted.
-default-members = [
-    ".",
     "crates/algebra",
     "crates/bivec",
     "crates/fp",
--- a/ext/Cargo.toml
+++ b/ext/Cargo.toml
@@ -63,10 +63,8 @@ nassau = []
 gpu = ["fp/gpu"]
 
 [workspace]
-# `fp-cuda` builds its CUDA kernel with nvcc, and fails to build without it. It is excluded from the
-# workspace rather than merely left out of `default-members`: the latter only affects commands that
-# omit `--workspace`, so `cargo check --workspace` and rust-analyzer's flycheck would still build
-# it, and fail for anyone without the CUDA Toolkit.
+# `fp-cuda` needs nvcc. `default-members` is not enough: `--workspace` selects every member, so
+# `cargo check --workspace` and rust-analyzer would still build it.
 exclude = ["crates/fp-cuda"]
 
 members = [
--- a/ext/build_docs.py
+++ b/ext/build_docs.py
@@ -16,9 +16,10 @@ EXCLUDED_FEATURES = {"default", "gpu"}
 
 
 def metadata(*args):
+    # Only stdout is captured, so cargo's diagnostics still reach the terminal if it fails.
     out = subprocess.run(
         ["cargo", "metadata", "--format-version", "1", *args],
-        capture_output=True,
+        stdout=subprocess.PIPE,
         check=True,
         text=True,
     )
@@ -45,11 +46,10 @@ def main():
 
     # Not `target/`: it moves when CARGO_TARGET_DIR is set.
     doc_dir = os.path.join(metadata()["target_directory"], "doc")
-    crates_js = os.path.join(doc_dir, "crates.js")
 
-    # Prevent the cached crates.js from confusing the current run.
-    if os.path.exists(crates_js):
-        os.remove(crates_js)
+    # The whole directory, not just a stale crates.js: CI caches it between runs and uploads it
+    # wholesale, so docs for a crate that is no longer documented would be published forever.
+    subprocess.run(["cargo", "clean", "--doc"], check=True)
 
     feature_args = ["--features", ",".join(features)]
     for argv in (
@@ -69,7 +69,7 @@ def main():
     # built from leaves no dangling entry for one that is excluded from the workspace.
     crates = sorted(p["name"].replace("-", "_") for p in packages if p["name"] != "ext")
     entries = ",".join(f"'{name}'" for name in [*crates, "ext"])
-    with open(crates_js, "w") as f:
+    with open(os.path.join(doc_dir, "crates.js"), "w") as f:
         f.write(f"window.ALL_CRATES = [{entries}];\n")
 
 
--- a/ext/build_docs.py
+++ b/ext/build_docs.py
@@ -16,7 +16,6 @@ EXCLUDED_FEATURES = {"default", "gpu"}
 
 
 def metadata(*args):
-    # Only stdout is captured, so cargo's diagnostics still reach the terminal if it fails.
     out = subprocess.run(
         ["cargo", "metadata", "--format-version", "1", *args],
         stdout=subprocess.PIPE,
--- a/ext/build_docs.py
+++ b/ext/build_docs.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+
+# Build the rustdoc output published to gh-pages.
+#
+# `--all-features` would enable `gpu`, which needs nvcc, so every other feature is named
+# explicitly. Both the feature list and the crate list are read from `cargo metadata`, so adding
+# a feature or a crate needs no change here.
+
+import json
+import os
+import subprocess
+import sys
+
+KATEX_HEADER = "gh-pages/katex-header.html"
+EXCLUDED_FEATURES = {"default", "gpu"}
+
+
+def metadata(*args):
+    out = subprocess.run(
+        ["cargo", "metadata", "--format-version", "1", *args],
+        capture_output=True,
+        check=True,
+        text=True,
+    )
+    return json.loads(out.stdout)
+
+
+def main():
+    meta = metadata("--no-deps")
+    packages = meta["packages"]
+
+    features = sorted(
+        f"{p['name']}/{feature}"
+        for p in packages
+        for feature in p["features"]
+        if feature not in EXCLUDED_FEATURES
+    )
+    if not features:
+        sys.exit("no features found; the metadata query is wrong")
+
+    env = dict(os.environ)
+    env["RUSTDOCFLAGS"] = (
+        f"--html-in-header {KATEX_HEADER} {env.get('RUSTDOCFLAGS', '')}".strip()
+    )
+
+    # Not `target/`: it moves when CARGO_TARGET_DIR is set.
+    doc_dir = os.path.join(metadata()["target_directory"], "doc")
+    crates_js = os.path.join(doc_dir, "crates.js")
+
+    # Prevent the cached crates.js from confusing the current run.
+    if os.path.exists(crates_js):
+        os.remove(crates_js)
+
+    feature_args = ["--features", ",".join(features)]
+    for argv in (
+        ["cargo", "rustdoc", "--examples", *feature_args],
+        [
+            "cargo",
+            "doc",
+            "--all",
+            "--no-deps",
+            "--document-private-items",
+            *feature_args,
+        ],
+    ):
+        subprocess.run(argv, check=True, env=env)
+
+    # Prevent the examples from showing up in the sidebar. Listing the same crates the docs were
+    # built from leaves no dangling entry for one that is excluded from the workspace.
+    crates = sorted(p["name"].replace("-", "_") for p in packages if p["name"] != "ext")
+    entries = ",".join(f"'{name}'" for name in [*crates, "ext"])
+    with open(crates_js, "w") as f:
+        f.write(f"window.ALL_CRATES = [{entries}];\n")
+
+
+if __name__ == "__main__":
+    main()
--- a/ext/crates/fp-cuda/README.md
+++ b/ext/crates/fp-cuda/README.md
@@ -63,8 +63,8 @@ at runtime, so the Rust side builds with no CUDA present.
 
 This crate is **excluded from the workspace**, so every workspace-wide command
 (`cargo build`, `cargo check --workspace`, `nix run .#test`, and rust-analyzer's
-flycheck) ignores it. It is opt-in: building requires nvcc on `PATH` and (at
-runtime) a Hopper-class GPU.
+flycheck) ignores it. It is opt-in: building requires nvcc on `PATH` and a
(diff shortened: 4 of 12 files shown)


ext/crates/fp-cuda/README.md
```diff
@@ -97,14 +97,11 @@ The kernel's tuning knobs live in `cuda_kernels/params.h`. The kernel includes
...
-`cargo check --workspace --all-features`, which the workspace CI runs on a
-machine with no CUDA. The stub carries no kernel, so at runtime
-`GpuContext::new` returns an error and callers (e.g. `fp`'s GPU dispatch)
-fall back to the CPU path. An nvcc that *is* present but fails to compile is
-still a hard build error.
+**When nvcc is absent** (CI, or a contributor without the CUDA Toolkit) the
+build fails. Nothing in the workspace's own `just` recipes builds this crate:
+`lint` and `test` pass `--exclude fp-cuda`, and it is not in `default-members`.
+Callers reach the backend through `fp`'s optional `gpu` feature, which is off
+by default and falls back to the CPU path when it is not enabled.
```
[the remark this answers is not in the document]
JoeyBF: Fixed in e5b70e979d — and it went further than flagged.

Both claims verified stale: `ext/justfile` has zero occurrences of `--exclude fp-cuda`, and `default-members` survives only as a prose mention inside a comment explaining why it *isn't* used.

Two more the review didn't catch:

- The "Why excluded from `default-members`?" section asserted "The crate is still a workspace **member**, so `cargo metadata` sees it, `rust-analyzer` indexes it" — the exact opposite of what this PR does, and the reason the change was needed.
- Six commands in the Prerequisites/Running sections used `-p fp-cuda` from the workspace root. Those now fail outright: `error: package ID specification `fp-cuda` did not match any packages`. They now run from `ext/crates/fp-cuda`.

.github/workflows/ext.yaml
```diff
@@ -112,6 +112,12 @@ jobs:
           toolchain: nightly
           components: clippy, rustfmt
 
+      # `just lint` drives clippy through `cargo hack --feature-powerset`.
+      - name: Install cargo-hack
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-hack
```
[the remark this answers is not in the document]
JoeyBF: Declining this one, though not because SHA pinning is wrong.

Every action in this workflow is referenced by tag, not digest:

```
actions/cache@v5          actions/checkout@v5
actions/download-artifact@v7  actions/setup-node@v5
actions/setup-python@v6   actions/upload-artifact@v6
dtolnay/rust-toolchain@v1 extractions/setup-just@v4
peaceiris/actions-gh-pages@v4
```

zizmor flags line 117 only because it is the line this PR adds; its blanket policy would flag all ten if the whole file were treated as new. Pinning the one action introduced by a PR about `cargo-hack` would leave the file inconsistent while improving the supply-chain posture only marginally — `taiki-e/install-action` would be pinned, the other nine still mutable.

Digest-pinning is a repo-wide decision worth making deliberately, ideally with something like Dependabot configured to bump the digests so they don't rot. Happy to open a separate issue for it if that's wanted; it shouldn't ride along here.

Same reasoning for the `excessive-permissions` warnings zizmor raised on lines 1-486 and 94-144: real, pre-existing, and workflow-wide rather than something this PR introduces.

.github/workflows/ext.yaml
```diff
@@ -112,6 +112,12 @@ jobs:
           toolchain: nightly
           components: clippy, rustfmt
 
+      # `just lint` drives clippy through `cargo hack --feature-powerset`.
```
hoodmane: Why is this comment here?

ext/crates/fp-cuda/src/lib.rs
```diff
@@ -62,8 +62,7 @@ pub struct GpuContext {
 impl GpuContext {
     /// Open device `device_id` and load the kernel onto it.
     ///
-    /// Fails if there is no usable device, or if the embedded PTX is the stub `build.rs` emits when
-    /// `nvcc` is absent, in which case the kernel is missing from the module.
+    /// Fails if there is no usable device.
```
hoodmane:
```suggestion
```

.github/workflows/ext.yaml
hoodmane:
```suggestion
```

ext/crates/fp-cuda/README.md
```diff
@@ -61,9 +61,10 @@ driver-API surface (module load, device buffers, typed launch) and its
 TMA descriptors. `cudarc` is stable Rust and dynamically loads the CUDA driver
 at runtime, so the Rust side builds with no CUDA present.
 
-This crate is **excluded from the workspace's `default-members`**, so plain
-`cargo build` / `nix run .#test` ignore it. It is opt-in: building requires
-nvcc on `PATH` and (at runtime) a Hopper-class GPU.
+This crate is **excluded from the workspace**, so every workspace-wide command
+(`cargo build`, `cargo check --workspace`, `nix run .#test`, and rust-analyzer's
+flycheck) ignores it. It is opt-in: building requires nvcc on `PATH` and (at
+runtime) a Hopper-class GPU.
```
hoodmane:
```suggestion
ignores it. It is opt-in: building requires nvcc on `PATH` and a Hopper-class GPU at runtime.
```

ext/examples/bruner.rs
```diff
@@ -168,6 +168,9 @@ fn read_bruner_resolution(data_dir: &Path, max_n: i32) -> Result<(i32, FiniteCha
     let cc = create_chain_complex(num_s as usize);
     let algebra = cc.algebra();
 
+    // With `nassau` the algebra is already a `MilnorAlgebra`, so this conversion is a no-op;
+    // without it the algebra is a `SteenrodAlgebra` and the conversion does real work.
```
hoodmane:
```suggestion
```

ext/examples/save_bruner.rs
```diff
@@ -18,6 +18,9 @@ fn main() -> anyhow::Result<()> {
 
     assert_eq!(resolution.prime(), 2);
     let algebra = resolution.algebra();
+    // With `nassau` the algebra is already a `MilnorAlgebra`, so this conversion is a no-op;
+    // without it the algebra is a `SteenrodAlgebra` and the conversion does real work.
```
hoodmane:
```suggestion
```

ext/Cargo.toml
```diff
@@ -63,23 +63,13 @@ nassau = []
 gpu = ["fp/gpu"]
 
 [workspace]
+# `fp-cuda` builds its CUDA kernel with nvcc, and fails to build without it. It is excluded from the
+# workspace rather than merely left out of `default-members`: the latter only affects commands that
+# omit `--workspace`, so `cargo check --workspace` and rust-analyzer's flycheck would still build
+# it, and fail for anyone without the CUDA Toolkit.
```
hoodmane: Make short.

ext/justfile
```diff
@@ -15,12 +15,10 @@ test:
     cargo test --features concurrent --doc --workspace
     pytest
 
+# `clippy` type-checks everything `check` does, so the powerset only needs the one pass.
```
hoodmane:
```suggestion
```

ext/justfile
```diff
@@ -30,12 +28,27 @@ fix:
     cargo clippy --workspace --all-targets --profile test --fix --allow-dirty --allow-staged
 
 docs:
+    #!/usr/bin/env bash
```
hoodmane: Let's make this a Python tool.

hoodmane (approved):
LGTM modulo comments.

ext/justfile
JoeyBF: Done in b2b619d3fb — `ext/build_docs.py`, following the `convert.py` convention (standalone `#!/usr/bin/env python3`, no packaging).

Worth flagging that this fixed a real bug rather than just reading better. The shell version used `jq`, which I never added to `flake.nix` — so `just docs` would have failed in a clean dev shell, and passed in CI only because the `ubuntu-latest` image happens to ship jq. The Python version parses `cargo metadata` with the stdlib, so the dependency is gone entirely.

Verified: `RUSTDOCFLAGS="-D warnings" just docs` exits 0, and `crates.js` comes out as `['algebra','bivec','fp','maybe_rayon','once','query','sseq','ext']` — matching what actually got documented, with no `fp_cuda` entry now that the crate is out of the workspace.

JoeyBF:
All nine comments addressed in b2b619d3fb.

**Comments removed** — workflow `cargo-hack` note, `GpuContext::new`'s "Fails if there is no usable device", and the `nassau` explanations above both `useless_conversion` attributes. The last two were especially redundant once you changed them to `#[cfg_attr(feature = "nassau", ...)]`, since the attribute already names the condition the comment was explaining.

**Shortened** — the `[workspace]` comment is now two lines, and the fp-cuda README sentence uses your phrasing.

**`build_docs.py`** — see the inline reply; it also removed an accidental `jq` dependency that was never in the flake.

Verified after the change: `just docs` exits 0 under `-D warnings`, `cargo fmt --check` clean, `black --check` clean on the new script.

Two open items on my side:

- The `CHANGES_REQUESTED` from CodeRabbit's first pass still sets `reviewDecision`, even though its follow-ups accepted both replies — those were posted as `COMMENTED`, which doesn't supersede it. Needs a re-review or a dismissal.
- `selenium (beta)` and `selenium (nightly)` failed on the previous run with an `api.snapcraft.io` HTTP 408 while installing Chromium, at the "Install dependencies" step before any test ran. Non-blocking (`continue-on-error` for non-stable) and unrelated to this branch, but it is #94 recurring.

ext/build_docs.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def metadata(*args):
+    out = subprocess.run(
+        ["cargo", "metadata", "--format-version", "1", *args],
+        capture_output=True,
+        check=True,
+        text=True,
+    )
+    return json.loads(out.stdout)
```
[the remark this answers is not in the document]
JoeyBF: Fixed in b65112685d. `metadata()` now captures only stdout, so cargo's own diagnostics reach the terminal. Verified: running it outside a cargo project prints `error: could not find \`Cargo.toml\`` above the traceback, where before you got only the traceback.

ext/build_docs.py
```diff
@@ -0,0 +1,77 @@
...
+        f"--html-in-header {KATEX_HEADER} {env.get('RUSTDOCFLAGS', '')}".strip()
+    )
+
+    # Not `target/`: it moves when CARGO_TARGET_DIR is set.
+    doc_dir = os.path.join(metadata()["target_directory"], "doc")
+    crates_js = os.path.join(doc_dir, "crates.js")
+
+    # Prevent the cached crates.js from confusing the current run.
+    if os.path.exists(crates_js):
+        os.remove(crates_js)
```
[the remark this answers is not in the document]
JoeyBF: Fixed in b65112685d — `cargo clean --doc` before the build, rather than removing just `crates.js`. This matters more than it looks on a PR that stops documenting a crate: the docs job caches `**/target` and uploads the directory wholesale, so cached `fp_cuda` docs would have kept being published after the crate left the workspace.

ext/justfile
```diff
@@ -17,10 +17,7 @@ test:
 
 lint:
     cargo fmt --all -- --check
-    cargo clippy --workspace --no-default-features --profile test
-    cargo clippy --workspace --all-targets --profile test
-    cargo check --workspace --no-default-features --profile test
-    cargo check --workspace --all-targets --all-features --profile test
+    cargo hack clippy --workspace --all-targets --feature-powerset --skip gpu --profile test
```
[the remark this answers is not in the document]
JoeyBF: Good catch, fixed in b65112685d by giving `fix` the same powerset rather than narrowing the comment. Narrowing would have reintroduced the exact blind spot this PR exists to close — `fix` silently covering less than `lint`. Verified `cargo hack` drives `clippy --fix` cleanly: exit 0 over all 82 combinations in 4m30s, with nothing left to change. It is slower than the old two invocations, but it is a developer-invoked command and the invariant now holds.

ext/crates/fp-cuda/README.md
```diff
@@ -61,9 +61,8 @@ driver-API surface (module load, device buffers, typed launch) and its
 TMA descriptors. `cudarc` is stable Rust and dynamically loads the CUDA driver
 at runtime, so the Rust side builds with no CUDA present.
 
-This crate is **excluded from the workspace's `default-members`**, so plain
-`cargo build` / `nix run .#test` ignore it. It is opt-in: building requires
-nvcc on `PATH` and (at runtime) a Hopper-class GPU.
+This crate is **excluded from the workspace**, so every workspace-wide command ignores it. It is
+opt-in: building requires nvcc on `PATH` and a Hopper-class GPU at runtime.
```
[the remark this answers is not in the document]
JoeyBF: Correct, fixed in 6c8a8fd100.

Verified rather than taken on faith — `cargo check --workspace --features gpu` does build the crate and fails on the missing nvcc, so "every workspace-wide command ignores it" was plainly wrong.

The README now says `--workspace` never *selects* it, and that it is built only as a path dependency when `fp`'s `gpu` feature is on. Two other places carried the same overclaim; one is fixed alongside, and the "Why excluded from the workspace?" section was already qualified by a following bullet stating the `gpu` path.

Also confirmed no `just` recipe enables `gpu` or `--all-features`, which is what makes the recipes safe on a machine without the toolkit.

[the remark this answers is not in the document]
JoeyBF: It's clear from context that we want to exclude fp-cuda by default, but allow enabling it when there is a GPU available.

PR #291, opened 2026-08-30
algebra: derive the free-module opgen table instead of storing it

JoeyBF:
`basis_element_to_opgen` stored one `OperationGeneratorPair` per basis element per degree. At high stems that table was roughly three quarters of the resolution's live heap, by a wide margin its largest single consumer.

All four of its fields are derivable. The basis of degree `t` is consecutive per-generator blocks in `(gen_deg, gen_idx)` order, and the block offsets are *already* stored in `generator_to_index`, which is indexed per generator and so is smaller than the per-basis-element table by a factor of the algebra's dimension. `index_to_op_gen` now does two binary searches — offsets to find the generator ordinal, `gen_deg_idx_to_internal_idx` to find that ordinal's degree — and computes `operation_index = index - offset` and `operation_degree = degree - generator_degree`.

What remains of the table is a per-degree dimension counter, which also takes over the mutex `add_generators` previously held on it.

Two subtleties worth knowing when reading the search:

- A generator with `num_ops == 0` yields an **empty** block that shares its offset with the next one, so the search must take the **last** block whose offset is `<= index`. Taking the first returns a zero-width block, and would be wrong exactly on the degrees where the algebra runs out of operations.
- `index_to_op_gen` now returns `OperationGeneratorPair` by value rather than by reference, since there is no longer a stored record to borrow. The type is `Copy`.

Callers walk their input indices in ascending order — the restricted-matrix builds do this per row — so consecutive lookups land in the same generator block nearly every time. A one-entry per-thread memo turns that common case into a range check rather than two searches. Its soundness rests on two things:

- Identity is a unique per-instance id from a global counter, **not** a pointer. An address can be reused after a module is dropped, and a stale hit would then answer from another module's layout.
- `generator_to_index` is append-only and existing offsets are never mutated, so a cached `[start, end)` stays valid for the module's life. For the last block, `end` is the dimension at cache time; if generators are added later the dimension grows and indices past the cached `end` simply miss and take the slow path. A hit therefore always describes the block it claims.

## Measurements

Against the stored-table version on a full-size resolution, comparing the two runs on the bidegrees they **both** computed — pairing on `(s, t)` and summing per-bidegree wall time, so the comparison controls for which bidegrees each run happened to reach:

| | |
|---|---|
| total work | unchanged, within 0.1% |
| per-bidegree wall time | median ratio 0.95 |
| peak RSS | a little over half the stored-table version's |

So the table can be dropped for roughly half the peak memory at no throughput cost.

These were measured on a downstream branch carrying the GPU resolution path, which is what makes resolutions of that size reachable; the algebra change itself is identical to the one here.

## Verification

`EXT_OPGEN_VERIFY=1` materialises the old table alongside and asserts every derived lookup, memo fast path included, against it. No disagreements over full resolutions at two different stems, and the resulting Ext charts are identical to those from the unmodified code. The flag should stay off outside validation — it restores the memory this removes.

`just lint` (including the `cargo hack` feature powerset), `just docs`, and `just test` all pass.

## Summary by CodeRabbit

* **Performance**
  * Improved free-module basis lookups, particularly for repeated and scattered access.
  * Improved free-module homomorphism operations through reusable lookup state.

* **New Features**
  * Added cursor-based access for efficient free-module traversal.
  * Added stateful iteration support for sequential and parallel execution.
  * Added binary-search helpers for initialized data collections.

* **Compatibility**
  * Updated basis lookup results for more convenient value-based use.

* **Reliability**
  * Improved handling of empty blocks, late generator additions, zero extensions, and odd-prime cases.

### DIFF
--- a/ext/crates/algebra/src/module/free_module.rs
+++ b/ext/crates/algebra/src/module/free_module.rs
@@ -1,4 +1,8 @@
-use std::sync::Arc;
+use core::range::Range;
+use std::sync::{
+    Arc,
+    atomic::{AtomicUsize, Ordering},
+};
 
 use fp::vector::{FpSlice, FpSliceMut};
 use once::{OnceBiVec, OnceVec};
@@ -8,7 +12,28 @@ use crate::{
     module::{Module, ZeroModule},
 };
 
-#[derive(Clone, Debug)]
+/// One generator's run of consecutive basis elements within a single degree.
+#[derive(Clone, Copy)]
+struct GeneratorBlock {
+    ordinal: usize,
+    generator_degree: i32,
+    generator_index: usize,
+    start: usize,
+}
+
+impl GeneratorBlock {
+    /// The pair for the basis element at `index`, which must lie in this block.
+    fn op_gen(&self, degree: i32, index: usize) -> OperationGeneratorPair {
+        OperationGeneratorPair {
+            generator_degree: self.generator_degree,
+            generator_index: self.generator_index,
+            operation_degree: degree - self.generator_degree,
+            operation_index: index - self.start,
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct OperationGeneratorPair {
     pub operation_degree: i32,
     pub operation_index: usize,
@@ -30,8 +55,18 @@ pub struct MuFreeModule<const U: bool, A: MuAlgebra<U>> {
     gen_names: OnceBiVec<Vec<String>>,
     /// degree -> internal index of first generator in degree
     gen_deg_idx_to_internal_idx: OnceBiVec<usize>,
+    /// internal index -> the degree of that generator.
+    ///
+    /// The inverse of `gen_deg_idx_to_internal_idx`, materialised so that
+    /// [`Self::index_to_op_gen`] can invert it by indexing rather than by searching. One `i32` per
+    /// generator, and there are far fewer generators than basis elements.
+    internal_idx_to_gen_deg: OnceVec<i32>,
     num_gens: OnceBiVec<usize>,
-    basis_element_to_opgen: OnceBiVec<OnceVec<OperationGeneratorPair>>,
+    /// degree -> number of basis elements.
+    ///
+    /// This is all that is kept of the per-basis-element table: the records themselves are derived
+    /// by [`Self::index_to_op_gen`].
+    dimensions: OnceBiVec<AtomicUsize>,
     /// degree -> internal_gen_idx -> the offset of the generator in degree
     generator_to_index: OnceBiVec<OnceVec<usize>>,
 }
@@ -43,6 +78,7 @@ impl<const U: bool, A: MuAlgebra<U>> std::fmt::Display for MuFreeModule<U, A> {
 }
 
 impl<const U: bool, A: MuAlgebra<U>> MuFreeModule<U, A> {
+    /// A free module with no generators yet; add them with [`Self::add_generators`].
     pub fn new(algebra: Arc<A>, name: String, min_degree: i32) -> Self {
         let gen_deg_idx_to_internal_idx = OnceBiVec::new(min_degree);
         gen_deg_idx_to_internal_idx.push(0);
@@ -52,8 +88,9 @@ impl<const U: bool, A: MuAlgebra<U>> MuFreeModule<U, A> {
             min_degree,
             gen_names: OnceBiVec::new(min_degree),
             gen_deg_idx_to_internal_idx,
+            internal_idx_to_gen_deg: OnceVec::new(),
             num_gens: OnceBiVec::new(min_degree),
-            basis_element_to_opgen: OnceBiVec::new(min_degree),
+            dimensions: OnceBiVec::new(min_degree),
             generator_to_index: OnceBiVec::new(min_degree),
         }
     }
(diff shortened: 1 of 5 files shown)


ext/crates/algebra/src/module/free_module.rs
```diff
@@ -338,9 +417,104 @@ impl<const U: bool, A: MuAlgebra<U>> MuFreeModule<U, A> {
...
+    /// generator, rather than from a record stored per basis element.
+    ///
+    /// Blocks may be empty: a generator admitting no operations in this degree shares its offset
+    /// with the next one. The search must therefore take the LAST block whose offset is at most
+    /// `index`, since taking the first would return a zero-width block.
+    pub fn index_to_op_gen(&self, degree: i32, index: usize) -> OperationGeneratorPair {
         assert!(degree >= self.min_degree);
-        &self.basis_element_to_opgen[degree][index]
+        debug_assert!(
+            index < self.dimensions[degree].load(Ordering::Acquire),
```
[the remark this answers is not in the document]
JoeyBF: Changed, though not for the stated reason.

The premise is not right: in release the lookup is already ordered without that load. `OnceVec::push` release-stores its `len` after writing the element, and `len()`/`Index` acquire-load it, so reading `generator_to_index[degree]` and the offsets inside it synchronizes with the pushes on its own. That is the guarantee the crate documents on `OnceVec::len`. The entries `add_generators` pushes into `internal_idx_to_gen_deg` and `gen_deg_idx_to_internal_idx` happen before the offset push in program order, so they are covered by the same edge. And a concurrently appended block cannot change an existing answer: new blocks start at offsets at or above the dimension the caller already observed, so `offset <= index` is false for every one of them and the search result is unchanged.

The load is now unconditional anyway, because the bounds check is worth having in release for its own sake. Without it an index past the dimension underflows `partition_point(..) - 1` and fails further in with an unrelated message. Two tests now pin that behaviour, one for an index past the dimension and one for a degree with no generators at all. It costs about 4 ns of the per-call form's 21 ns, and nothing on the cursor path that the dense callers use.

ext/crates/algebra/src/module/homomorphism/free_module_homomorphism.rs
```diff
@@ -60,30 +102,55 @@ where
...
+        let source = &*self.source;
+        matrix.maybe_par_iter_mut().enumerate().for_each_init(
+            || source.opgen_cursor(degree),
+            |cursor, (i, row)| self.apply_to_basis_element_with(cursor, row, 1, i),
+        );
+    }
+
+    /// See [`Self::get_matrix`] for the cursor.
+    fn get_partial_matrix(&self, degree: i32, inputs: &[usize]) -> Matrix {
+        let mut matrix = Matrix::new(self.prime(), inputs.len(), self.target.dimension(degree));
```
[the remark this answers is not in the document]
JoeyBF: Applied.

Worth noting for anyone reading the diff that this is not a regression from this PR: the default `get_partial_matrix` in `homomorphism/mod.rs` has the same line, and the override copied it faithfully. It is latent rather than live, since every current caller (`nassau.rs` and `sq0.rs`) goes through a homomorphism whose shift is zero, which is why nothing failed. But the sibling `get_matrix` asserts the shifted dimension, so the two disagreed, and the override is the one place where that mattered enough to fix.

Left the trait default alone. The same one-line change there would be a no-op for every caller in the tree today, but it changes shared behaviour for any implementor, and that belongs in its own change rather than buried in this one.

ext/crates/algebra/src/module/free_module.rs
```diff
@@ -391,6 +495,65 @@ pub struct GeneratorData<const N: usize> {
...
+/// [`MuFreeModule::index_to_op_gen`] searches the block offsets on every call. A caller walking a
+/// degree's basis crosses a block boundary only once every `num_ops` indices, so remembering the
+/// current block answers almost every lookup with a range check instead.
+///
+/// The cursor borrows the module, which is what makes it sound: it cannot outlive the module whose
+/// layout it caches, and `generator_to_index` is append-only, so the block it remembers is never
+/// rewritten. It is also pinned to the dimension `degree` had when it was created, which is what
+/// bounds the last block -- that block is the only one with no following offset to end it, and a
+/// concurrent [`MuFreeModule::add_generators`] would otherwise extend it under the cursor.
+///
```
hoodmane, line 498:
```suggestion
```

ext/crates/algebra/src/module/free_module.rs
```diff
@@ -391,6 +495,65 @@ pub struct GeneratorData<const N: usize> {
...
+/// [`MuFreeModule::index_to_op_gen`] searches the block offsets on every call. A caller walking a
+/// degree's basis crosses a block boundary only once every `num_ops` indices, so remembering the
+/// current block answers almost every lookup with a range check instead.
+///
+/// The cursor borrows the module, which is what makes it sound: it cannot outlive the module whose
+/// layout it caches, and `generator_to_index` is append-only, so the block it remembers is never
+/// rewritten. It is also pinned to the dimension `degree` had when it was created, which is what
+/// bounds the last block -- that block is the only one with no following offset to end it, and a
+/// concurrent [`MuFreeModule::add_generators`] would otherwise extend it under the cursor.
```
hoodmane:
```suggestion
```

ext/crates/algebra/src/module/free_module.rs
```diff
@@ -391,6 +495,65 @@ pub struct GeneratorData<const N: usize> {
     pub end: [usize; N],
 }
 
+/// A position in one degree's basis, for callers that look up many indices in ascending order.
+///
+/// [`MuFreeModule::index_to_op_gen`] searches the block offsets on every call. A caller walking a
```
hoodmane:
```suggestion
/// A caller walking a
```

ext/crates/algebra/src/module/free_module.rs
```diff
@@ -391,6 +495,65 @@ pub struct GeneratorData<const N: usize> {
...
+
+impl<const U: bool, A: MuAlgebra<U>> OpGenCursor<'_, U, A> {
+    /// The `(operation, generator)` pair for basis element `index` of this cursor's degree.
+    pub fn get(&mut self, index: usize) -> OperationGeneratorPair {
+        assert!(
+            index < self.dimension,
+            "index {index} is past the dimension this cursor was created with"
+        );
+        if !self.range.contains(&index) {
+            self.block = self.module.block_containing(self.degree, index);
```
hoodmane, line 522: Maybe put a comment that this does a full binary search every time we move blocks, whereas we could later improve this by doing a linear scan from the current block, particularly if index is only one outside of the current range. Or maybe we should just fix it?

hoodmane (approved):

ext/crates/algebra/src/module/free_module.rs
hoodmane: I guess there are two cases: One is when we are doing an actual linear scan which would be better served by an actual iterator that knows it will go over these in order. The other case is applying a algebra element to a module element where we know that all the outputs from each basis element of the module will end up in the same block so we will see many accesses to the same block in a row but when we switch blocks the new block is unpredictable. Could you add a comment explaining this expected access pattern?

PR #266, opened 2026-07-12
Motivic foundation: A_C, the C-motivic Steenrod algebra engine over F₂[τ]

JoeyBF:
This is the **foundation layer** of a larger effort to compute the C-motivic Adams E₂ page by deformation. The full work is being split into a stack of independently reviewable PRs; this one has no dependencies and touches only the `algebra` crate.

## What it adds

- **`motivic::tau`** — the coefficient ring F₂[τ] as a small homogeneous scalar (`Tau`). Everything in the motivic world is weight-homogeneous, so every structure constant is a *single* power of τ. `Tau` encodes that as a one-integer valuation (not a polynomial), and carries the whole τ-tower so it never has to be threaded through the resolution engine.

- **`motivic::milnor`** — `MotivicMilnorAlgebra`, i.e. A_C presented as a free F₂[τ]-module on the Milnor basis Q(E)P(R). The product is computed **two independent ways**:
  - a *duality oracle* — dualize the coproduct ψ of the dual algebra A_∗∗ (correctness-oriented, obviously-correct);
  - the *closed-form* product of Kong–Lin, *Product formulas for motivic Milnor basis* ([arXiv:2411.12890](https://arxiv.org/abs/2411.12890)), Theorem 5.1, specialized to the C-motivic point ρ = 0 (fast).

  The fast path is validated exhaustively against the oracle over a range. This is deliberately **not** an `Algebra` impl — that trait is over F_p; this is the F₂[τ] *engine* the rest of the deformation builds on.

## What it deliberately does *not* do

Nothing here wires into the resolution engine. The mod-τ reduction A_C/τ (the connected finite-type F₂-algebra the engine actually resolves, giving the algebraic Novikov E₂) is a small `Algebra` impl layered on top of this engine in the **next** PR, so this one stays a self-contained, heavily-tested combinatorial core.

## Docs

`MOTIVIC_GPU_HANDOFF.md` documents the product engine's batching boundary for a future GPU port.

## Tests

`rewrite_tau` identities, product associativity, weight-homogeneity (product weight − τ-power is additive), and closed-form-vs-duality agreement across a bidegree range. `cargo test -p algebra --lib motivic` — 22 passing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_013ePtYD7Bt4iPeCtmqtqvZE)_

## Summary by CodeRabbit

* **New Features**
[...]

PR #292, opened 2026-08-30
algebra: extract the Milnor product and derive `generic` from the prime

JoeyBF:
Pure refactor of the Milnor product path — no behaviour change, no new API surface beyond one free function.

### Extract `milnor_product`

`multiply_with_allocation` branched on `generic()`, but the two arms were the same computation. `multiply_qpart(m1, 0)` returns `[(1, m1)]` — `BitflagIterator::set_bit_iterator(0)` yields nothing — so the classical arm is the generic arm with an empty exterior part. The two are collapsed into one free function:

```rust
pub fn milnor_product(
    p: ValidPrime,
    m1: MilnorBasisElement,
    m2: MilnorBasisElement,
    allocation: PPartAllocation,
    f: impl FnMut(u32, &MilnorBasisElement),
) -> PPartAllocation
```

which reports each `(coefficient, basis element)` through a callback, leaving the index lookup and excess filter to the caller. `multiply_qpart` becomes free too; it only ever read `self.p`. `MilnorAlgebra::multiply_with_allocation` is now a thin wrapper.

The motivation is downstream: $A_C/\tau \cong \mathbb{F}_2[\xi_i] \otimes E(\tau_i)$ has the odd-primary dual's shape with $2^i$ in place of $p^i$, so its product is the same commutation rule plus one `PPartMultiplier` walk. Sharing this function rather than reimplementing it keeps one algorithm under test instead of two.

### Derive `generic` from the prime

The field was written once by the constructor as `p != 2` and never mutated — a cached derivation, not state. `generic()` now computes it.

That also drops its `#[cfg(feature = "odd-primes")]` gate. Without the feature `ValidPrime` is the zero-sized type $2$, so `self.p != 2` folds to a constant `false` on its own; the cfg was doing nothing the type system wasn't already doing.

Removing the field also closes a latent trap. A hand-constructed `p = 2, generic = true` algebra has exactly the graded dimensions of $A_C/\tau$ (41 degrees checked, 0 differ), which makes it tempting as a shortcut — but `generators()` finds $Q_k$ via `factor_pk(p, degree + 1) == (k, 2)`, the odd-primary $|Q_k| = 2p^k - 1$. At $p = 2$, `factor_pk(2, 2^{k+1}) = (k+1, 1)`, so every $Q_k$ past $Q_0$ silently vanishes from the generating set. Correct at odd primes, unreachable through the constructor today, and now unreachable by construction.

### Checks

`cargo test -p algebra` (63 pass), `cargo fmt --check`, `cargo clippy --all-targets`, `cargo check --no-default-features`, and `just docs` under CI's `RUSTFLAGS`/`RUSTDOCFLAGS: -D warnings` — all clean.

### Relationship to the motivic work

Independent of #266 — the two touch disjoint regions of `milnor_algebra.rs` and rebase past each other in either order. Only the follow-up $A_C/\tau$ PR needs both.

https://claude.ai/code/session_01F25ZVbsP7ULg41iY3MP6FX

## Summary by CodeRabbit

* **New Features**
  * Added a public Milnor product operation for multiplying Milnor basis elements with allocation support.

* **Improvements**
  * Unified multiplication behavior across supported prime configurations.
  * Improved handling of exterior and polynomial components during multiplication.

### DIFF
--- a/ext/crates/algebra/src/algebra/milnor_algebra.rs
+++ b/ext/crates/algebra/src/algebra/milnor_algebra.rs
@@ -414,8 +414,6 @@ impl std::fmt::Display for MilnorBasisElement {
 pub struct MilnorAlgebra {
     profile: MilnorProfile,
     p: ValidPrime,
-    #[cfg(feature = "odd-primes")]
-    generic: bool,
 
     unstable_enabled: bool,
 
@@ -455,8 +453,6 @@ impl MilnorAlgebra {
         assert!(profile.is_valid());
         Self {
             p,
-            #[cfg(feature = "odd-primes")]
-            generic: p != 2,
             unstable_enabled,
             profile,
             ppart_table: OnceVec::new(),
@@ -468,17 +464,13 @@ impl MilnorAlgebra {
         }
     }
 
+    /// Whether the algebra has an exterior part, which is the case exactly at odd primes.
+    ///
+    /// Without the `odd-primes` feature [`ValidPrime`] is the zero-sized type $2$, so this folds
+    /// to a constant `false`.
     #[inline]
     pub fn generic(&self) -> bool {
-        #[cfg(feature = "odd-primes")]
-        {
-            self.generic
-        }
-
-        #[cfg(not(feature = "odd-primes"))]
-        {
-            false
-        }
+        self.p != 2
     }
 
     pub fn q(&self) -> i32 {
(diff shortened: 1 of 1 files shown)


JoeyBF:
@coderabbitai review

JoeyBF:
Closing in favor of #293

PR #298, opened 2026-09-19
fp-cuda: compile the kernel at runtime with NVRTC

JoeyBF:
`fp-cuda`'s `build.rs` shelled out to nvcc to produce `matmul_b1.ptx`, which `src/lib.rs` embedded. cudarc already links NVRTC, so the kernel source is embedded instead and compiled on the first `GpuContext::new` (~0.2 s, cached per process — once per source+arch, not once per device). The driver JITs the PTX exactly as before.

`build.rs` and `cuda_kernels/params.h` are both gone (176 lines), and with them the `#define <NAME> <integer>` scraper, the generated Rust mirror, and the `FP_CUDA_VERIFY` static_assert block whose only job was to catch that scraper misreading the header. NVRTC takes `-D` options, so the knobs live in `src/params.rs` and go to the kernel from there: one set of values, nothing to parse, nothing to cross-check. `matmul_b1.cu` defines no knob itself and `#error`s if one is missing.

The kernel needed two things from the headers NVRTC does not have — the stdint typedefs and `CUtensorMap`, which it only ever takes the address of. Both are declared under `__CUDACC_RTC__`; the `#else` branch keeps the file compilable by nvcc directly, for reading `ptxas -v` register counts off it.

### Consequences

- Building needs no CUDA Toolkit at all, so the crate rejoins the workspace: `cargo check --workspace`, `just lint` and rust-analyzer now cover its host code. Running needs libnvrtc in addition to the driver; the `#gpu` nix shell already provides it, and `GpuContext::new` returns `Err` (CPU fallback) when it is absent rather than panicking.
- `--skip gpu` is dropped from `just lint`/`just fix`, since it was there because building `fp-cuda` needed nvcc. Clippy now sees `fp::blas`'s dispatch arm and the `cuda_dispatch` integration test, which was `#[cfg(feature = "gpu")]` and so entirely invisible before. The feature powerset doubles in size, which is the cost of covering the axis at all.
- A kernel that does not compile is caught by `cargo test` rather than `cargo build`: the new `kernel_compiles` test needs no GPU, only the toolkit, which is the same condition under which the nvcc step used to run.
- `examples/kernel_ptx.rs` prints the generated PTX for `ptxas -v` inspection.

### Testing

Verified against NVRTC 12.9: `matmul_b1.cu` compiles for `compute_90a`, and the resulting PTX assembles under `ptxas -arch=sm_90a -O3` with no spills — 254 registers for `matmul_b1_kernel`, 32 for `transpose_tile_b1_kernel`. The missing-knob guard was exercised by dropping `GROUP_M` from `params::defines()`; NVRTC's log names the file, line and required flags. `cargo clippy` is clean for `fp-cuda` and for `fp` under `gpu`, `gpu,concurrent`, `--no-default-features --features gpu`, and `--no-default-features --features gpu,proptest`.

**Not yet run on a Hopper device** — `cargo run -p fp-cuda --example matmul_b1_demo` in the `#gpu` shell is still the gate on correctness.

One observation, unrelated to this change but visible from the new example: `ptxas` reports `(C7508) 'setmaxnreg' ignored; unable to determine register count at entry` on the standalone PTX. That comes from assembling PTX without a pinned entry register count, not from NVRTC — the source is byte-identical to what nvcc `-ptx` was fed, and the driver JIT was already doing this — but it may be worth confirming the producer/consumer register split is taking effect.

https://claude.ai/code/session_01QARTbbPgHBojQ1vYX9FBHR

---
_Generated by [Claude Code](https://claude.ai/code/session_01QARTbbPgHBojQ1vYX9FBHR)_

## Summary by CodeRabbit

* **New Features**
  * GPU support can now be built without the CUDA Toolkit or `nvcc`.
  * CUDA kernels compile at runtime through NVRTC and are cached for reuse.
  * Added an example for generating and inspecting compiled PTX.
* **Bug Fixes**
  * When NVRTC is unavailable, GPU initialization reports an error and GPU-enabled execution falls back to CPU processing.
* **Documentation**
  * Updated CUDA setup, runtime requirements, tuning guidance, and workspace usage documentation.


ext/crates/fp-cuda/cuda_kernels/matmul_b1.cu
```diff
@@ -25,11 +25,28 @@
...
+
+#ifdef __CUDACC_RTC__
+// NVRTC compiles a bare string with no filesystem behind it, so no headers are available. Every
+// other CUDA name here is an NVRTC builtin.
+using int32_t  = int;
+using uint32_t = unsigned int;
+using uint64_t = unsigned long long;
+// Opaque stand-in for the driver's CUtensorMap, laid out as CUtensorMap_st is. The kernel only
+// takes its address, so layout compatibility is all the parameter needs.
+struct alignas(64) CUtensorMap { uint64_t opaque[16]; };
```
JoeyBF, line 43: This is the line in the diff most worth scrutinising, since it hand-rolls a driver ABI type.

The driver's definition is `typedef struct CUtensorMap_st { alignas(64) cuuint64_t opaque[16]; } CUtensorMap;` — 128 bytes, 64-byte aligned, which is what this matches. It is safe to restate only because the kernel never reads a field: `tma_2d` and `tma_store_2d` take `const CUtensorMap*` and pass the address as an `"l"` operand to `cp.async.bulk.tensor.*`. Layout compatibility with the by-value `__grid_constant__` parameter is the whole contract.

Verified end to end against NVRTC 12.9: the kernel compiles for `compute_90a` and the resulting PTX assembles under `ptxas -arch=sm_90a -O3` with no spills (254 registers for `matmul_b1_kernel`, 32 for `transpose_tile_b1_kernel`), i.e. the descriptor argument round-trips.

---

ext/crates/fp-cuda/src/lib.rs
```diff
@@ -4,29 +4,74 @@ use std::{
...
+const ARCH: &str = "compute_90a";
+
+/// Compile the kernel to PTX with NVRTC, passing the knobs from [`params::defines`].
+///
+/// Needs `libnvrtc`, but no GPU.
+pub fn compile_kernel() -> anyhow::Result<Ptx> {
+    // SAFETY: `is_culib_present` only `dlopen`s the candidate library names and reports whether
+    // one resolved; it dereferences nothing. cudarc panics on the first missing symbol, so the
+    // probe has to come before any other nvrtc call.
+    if !unsafe { cudarc::nvrtc::sys::is_culib_present() } {
```
JoeyBF, line 40: Why this probe is here rather than letting the call fail: cudarc's `dynamic-loading` resolves `libnvrtc` lazily and `panic!`s the first time it reaches for a symbol in a library that is not there (`get_lib_name_candidates` → `panic_no_lib_found`). Without the probe a toolkit-less machine would abort instead of returning `Err`, and `fp::blas::cuda::context()` — which is `GpuContext::new(0).ok()` — would never get to fall back to the CPU path.

The same probe guards the `kernel_compiles` test, which is why that test passes trivially on CI rather than failing there.

---

ext/crates/fp-cuda/src/lib.rs
```diff
@@ -4,29 +4,74 @@ use std::{
...
+        arch: Some(ARCH),
+        // Names the program, so NVRTC's diagnostics say matmul_b1.cu rather than "default_program".
+        name: Some("matmul_b1.cu".to_string()),
+        options: params::defines(),
+        ..Default::default()
+    };
+
+    compile_ptx_with_opts(KERNEL_SRC, opts).map_err(|e| match &e {
+        // `CompileError`'s `Display` is a `Debug` dump that buries the log in escapes.
+        CompileError::CompileError { log, .. } => anyhow!(
```
JoeyBF, line 58: This arm is what makes moving compilation to runtime tolerable, so it is worth seeing the output it produces. Dropping `GROUP_M` from `params::defines()` gives:

```
matmul_b1.cu must compile: NVRTC failed to compile matmul_b1.cu for compute_90a:
matmul_b1.cu(33): catastrophic error: #error directive: "compile this with -DMSTRIPS=.. -DMW=.. ...; see src/params.rs"
  #error "compile this with -DMSTRIPS=.. -DMW=.. -DSTAGES=.. -DGROUP_M=.. \
   ^
1 catastrophic error detected in the compilation of "matmul_b1.cu".
```

File, line and the required flags. Via `CompileError`'s own `Display` that same log arrives as a `Debug` dump with the newlines escaped, which is unreadable in a test failure.

---

ext/justfile
```diff
@@ -17,12 +17,12 @@ test:
 
 lint:
     cargo fmt --all -- --check
-    cargo hack clippy --workspace --all-targets --feature-powerset --skip gpu --profile test
+    cargo hack clippy --workspace --all-targets --feature-powerset --profile test
```
JoeyBF, line 20: Dropping `--skip gpu` doubles the powerset (`ext` has six features, so 32 → 64 combinations) and `lint` is already the long pole — it took ~6 min on this PR's runs. The gain is that clippy now sees `fp::blas`'s dispatch arm and `fp/tests/cuda_dispatch.rs`, which is `#[cfg(feature = "gpu")]` and was therefore invisible to every previous lint run.

If that trade is not worth it, `--group-features` over the gpu axis would keep most of the coverage at half the combinations. Happy to change it either way.

---

JoeyBF (commented):
Notes for review, kept out of the tree because they stop being useful once this merges.

Two comments in `lib.rs` carried their own justification; `e5f74a3` cuts them back to the part a maintainer still needs — that cudarc panics on a missing symbol, and that `CompileError`'s `Display` is a `Debug` dump. The reasoning behind both is inline below instead.

Still unverified: nothing here has run on a Hopper device. `cargo run -p fp-cuda --example matmul_b1_demo` in `nix develop ./ext#gpu` is the gate on correctness.

---

